PR #79, opened 2014-12-29
[DONT MERGE] Make TwitterSearchSensor rate-limit aware

Kami:
Note: This should be merged when StackStorm v0.7.0 is out.

If a rate limit has been reached, we will now automatically increase poll
intervall and restore it back to the original poll interval when requests
are not throttled anymore.

Note: Rate limit should only be reached if user uses really low poll interval
(e.g. 0.2 seconds) or uses same credentials for multiple applications.


Kami:
Closing this - with the new code changes is basically impossible to hit the rate limit unless you are using the same access token in 20 different apps.

PR #87, opened 2014-12-31
[WIP] Add script which verifies pack contains a README which satisfies a particular criteria

Kami:
This pull request adds a "lint" script which checks that pack contains a "README.md" file which satisfies the following criteria:
- README.md file exists
- Readme contains configuration section if config.yaml file exists
- Readme contains a section for each sensor available in the pack
- Readme contains a section for each action available in the pack

Note 1: Yes, the script is a bit hacky, but automated checks are better and less error-prone than no or manual checks. And the old "no tests and documentation means it doesn't exist" phrase still stands.

Note 2: Yes, the script enforces a particular README structure, but that's not bad, in fact, quite the contrary. Structure and consistency is good. Eventually, we should also create a "pack bootstrap tool" which creates a README.md template which fits this particular structure.

### DIFF
--- a/scripts/validate-pack-readme.py
+++ b/scripts/validate-pack-readme.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python
+
+import os
+import re
+import glob
+import json
+import argparse
+
+import yaml
+
+"""
+Script which verifies that pack contains README which meets the following criteria:
+
+1. Readme contains configuration section if config.yaml file exists
+2. Readme contains a section for each of the available actions and sensors
+"""
+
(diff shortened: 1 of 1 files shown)


Kami:
I'll close this for now and re-open it against new exchange repo in the future.

emedvedev:
That's nice, although we'll need to document this structure very clearly, and ideally have an boilerplate before we enforce. I also like the approach used in [godoc](https://godoc.org): if your code is commented properly, it will compile a nice manual with examples, method lists, and comments as help strings. I wonder if we can pull this off in the future, even if in a very minimal version.

PR #88, opened 2015-01-01
Cleaning up after last PR

DoriftoShoes:
This is in response to some feedback @Kami gave me on the last PR.

* Removed _init__
* Added spaces after commas
* Removed spaces in assignments
* Added a log line and False return if retries are exceeded.


Kami:
LGTM, +1

PR #89, opened 2015-01-06
Add clone action to Git

jfryman:
This PR adds the ability for StackStorm to download a repository to a remote machine for manipulation. Used in the CI/CD workflow.

### DIFF
--- a/packs/git/actions/clone.yaml
+++ b/packs/git/actions/clone.yaml
@@ -1,0 +1,25 @@
+---
+name: clone
+runner_type: run-remote
+description: "Clone a repository"
+enabled: true
+entry_point: ''
+parameters:
+  source:
+    type: string
+    description: Source URI to clone a repository from
+    required: true
+    position: 1
+  ref:
+    type: string
+    description: Requested git ref to check the repository out to.
+    default: master
+    position: 2
+  destination:
+    type: string
+    description: Location on the filesystem to clone the repository to
+    required: true
+    position: 3
+  cmd:
+    default: "git clone -b {{ref}} {{source}} {{destination}}"
+    immutable: true
--- a/packs/git/config.yaml
+++ b/packs/git/config.yaml
@@ -1,6 +1,2 @@
 ---
   url: "git@demo-git:/home/git/repos/test.git"
-
-  # For Web-Based Requests
-  user: ""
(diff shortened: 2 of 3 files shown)


packs/git/config.yaml
```diff
@@ -1,2 +1,6 @@
 ---
   url: "git@demo-git:/home/git/repos/test.git"
+
+  # For Web-Based Requests
+  user: ""
```
Kami: Hm, those two settings seems to be a leftover from a github pack and are not really used here or?

Kami:
New action LGTM, +1

packs/git/rules/github-pr-merge.json
```diff
@@ -9,16 +9,7 @@
         }
     },
 
-    "criteria": {
```
Kami: I assume you have intentionally removed this section?

It looks like previously we sent all the github events to this webhook and now we only send merge events to this webook?

jfryman: This one I wasn't sure about. I got this change on a merge from master... it's easily revert-able. I'm not doing anything with this.

packs/git/config.yaml
jfryman: Sure is. :blush:. Let me get this out too. :)

packs/git/actions/clone.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: clone
+runner_type: run-remote
+description: "Clone a repository"
+enabled: true
+entry_point: ''
+parameters:
+  source:
```
manasdk: I guess this only works for oublic repos unless the system user has access to these repos - right? 

I am ok with this assumption but wanted to clarify if that is your intent and how it works.

jfryman: Nope, I'm using it for a private repository. It just requires that SSH Keys exist on the system. This is not how I envision the end-state to be, but it's Good Enough(tm) for now.

I would like to be able to add flags like AUTH or SSH_KEY without having to write a secondary script to handle the parameter passing and assembling the final command.

manasdk: Yeah being able to provide location to a key or auth creds would really help. I agree this works for now.

PR #90, opened 2015-01-07
Pwh/dig action

DoriftoShoes:
This adds a dig action.

The supported parameters are:
* nameserver - the nameserver to query.  Defaults to 127.0.0.1
* hostname - the hostname to lookup
* queryopts - Dig query options.  Defaults to short
    * See http://linux.die.net/man/1/dig
* count - number of results to return
* rand - this randomizes the returned list

### DIFF
--- a/packs/linux/actions/dig.py
+++ b/packs/linux/actions/dig.py
@@ -1,0 +1,29 @@
+#! /usr/bin/python
+
+import subprocess, random, re
+from st2actions.runners.pythonrunner import Action
+
+class DigAction(Action):
+
+    def run(self,rand,count,nameserver,hostname,queryopts):
+        opt_list = []
+        output = []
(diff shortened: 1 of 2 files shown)


Kami:
I know you already merged it, but I'll still add some stylistic comments (hope you don't mind) :P

packs/linux/actions/dig.py
```diff
@@ -0,0 +1,29 @@
...
+
+class DigAction(Action):
+
+    def run(self,rand,count,nameserver,hostname,queryopts):
+        opt_list = []
+        output = []
+        nameserver = '@' + nameserver
+        cmd_args = ['dig', nameserver]
+
+        if re.search(',',queryopts):
```
Kami: In this case you don't really need to use regular expression, you could just do ``if ',' in queryopts`.

PR #94, opened 2015-01-18
Add new FileWatch sensor which monitors files for activity (new lines)

Kami:
This sensor monitors files for activity and emits a trigger every time a new line is detected.

This sensor allows users to do many cool things:

* Ship log lines to logging services (loggly, logstash, splunk, etc.)
* Implement tools such as fail2ban on top of StackStorm
* ...

Underneath it depends on logshipper which uses wrapped version of pyinotify which cooperates with eventlet.

I've put the sensor in a linux pack, but if you think it belongs in a separate pack and you have a suggestion for a pack name, let me know.

### DIFF
--- a/packs/linux/README.md
+++ b/packs/linux/README.md
@@ -2,6 +2,29 @@
 
 This pack contains actions for commonly used Linux commands and tools.
 
+## Configuration
+
+* ``file_watch_sensor.file_paths`` - A list of paths to files to monitor.
+
+## Sensors
+
+### FileWatchSensor
+
+This sensor monitors specified files for new new lines. Once a new line is
+detected, a trigger is emitted.
+
+### linux.file_watch.line trigger
+
+Example trigger payload:
+
+```json
+{
+    "file_path": "/var/log/auth.log",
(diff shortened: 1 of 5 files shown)


jfryman:
This is cool. But, it'll only run on a system where a worker executes, correct?

Probably not for thought on this PR, but in the larger design state: I'd really love to be able to stick this on a few hosts with an agent to use it as a slurper of sorts... how could I achieve this?

Kami:
Yep - only on the host where the sensor container and sensor is running. Other sensors have similar "limitation".

It's something we have discussed in the past (load balancing and running sensors on multiple servers) and we have made some progress towards it (via the sensor wrapper), but we haven't agreed on a full and final solution just yet.

As far as running this on multiple servers right now goes - we could modify the sensor / script to submit results (triggers) via the API (webhook trigger).

Kami:
For a short term solution - we could experiment with writing a small wrapper script, which runs a sensor and submits results via the API instead of via the message bus (this shouldn't be too hard because of the existing sensor wrapper).

PR #96, opened 2015-01-20
traceroute pack

Aamir-raza-1:
Traceroute pack initially will accept 3 arguements. One is mandatory that is the host to traceroute, the rest of the two are optional.  Limiting the max number of  hops, Number of queries to each hop.
default command:
traceroute google.com
Passing arguements:
traceroute 10 3 google.com

if Default arguments are not passed it will pick up default

### DIFF
--- a/packs/traceroute/actions/traceroute.sh
+++ b/packs/traceroute/actions/traceroute.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+HOST=$1
+MAX_HOPS=$2
+MAX_QUERIES_TO_HOP=$3
+
+MAX_HOPS="${MAX_HOPS:-30}"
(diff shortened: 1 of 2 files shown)


Kami:
Great, thanks.

As far as the pack name goes - It looks like this action is Linux specific so it's probably better to move it to the existing `linux` pack (https://github.com/StackStorm/st2contrib/tree/master/packs/linux).

manasdk:
+1 to moving it to linux pack over it own pack.

Kami:
I've moved action into linux pack and merged changes into master. Thanks.

On top of that, I also changed the action to use remote runner (just thought about that when moving the action). Previously it was using a local runner which means the action would only run on the same host as the action runner service. Now the action can run on any host to which action runner has access.

Aamir-raza-1:
ok, thats great !! Thanks

PR #99, opened 2015-01-25
scanning open ports

Aamir-raza-1:
PortScanner:
This pack takes Host IP as an argument and scans for the open ports using nmap. It lists the Open ports, thier protocol type(tcp/udp etc), thier status(open, filtered etc) and the service which is running on open ports. It takes one argument that is the host IP, which is mandatory
Usage:
st2 run get_open_ports.get_open_ports host=<1.1.1.1> # give the host for scanning opened ports


Kami:
Thanks.

Same as with the traceroute action, this one should also live in the `linux` pack.

Kami:
Also need to add `python-nmap` dependency to `linux/requirement.txt` when you move the action.

Aamir-raza-1:
Kami can you please guide me a little about how can i move this pack into linux. Thanks, appreciate your kind and helping attitude

Kami:
@Aamir-raza-1 sure, just move the action file (get_open_ports.py) and action metadata file (get_open_ports.yaml) to `packs/linux/actions/` directory.

After that, you can also do `git rm -r packs/get_open_ports/`

Aamir-raza-1:
Great, Thanks !!

Aamir-raza-1:
@Kami , Done !! Thanks

Kami:
@Aamir-raza-1 Thanks. I've made some changes and merged PR into master:

1. Add missing python-nmap dependency to requirements file - 5b881d2d274d7a63fa8bee359df08eb2b8cfffbb
2. Removed obsolete main entry point - eb2f19a95175d68c5ac5345d38c8ce8db3b3ba9c
3. Added a note about nmap and root dependency - 2a99d8e3877162a3e06b324b84cb92ae26661d53

We still need to decide how to handle root requirement better. Currently, Python actions run as unprivileged user (stanley) which means this action won't work. One of the solutions would be to drop flags which require root privileges (-sU - UDP ports scan, -sZ).

/cc @DoriftoShoes @jfryman What do others think?

jfryman:
> We still need to decide how to handle root requirement better. Currently, Python actions run as unprivileged user (stanley) which means this action won't work. One of the solutions would be to drop flags which require root privileges (-sU - UDP ports scan, -sZ).

Yes, they need to be dropped, imo. I think that our goal should be to not require users to modify the integration point beyond an action connection. In this case, we'd need `nmap` SUID, which I see as a no-go for a lot of places.

I think this is good that we discovered it. Thanks @Aamir-raza-1 @Kami. :+1: In general though, this feels like a pattern we should capture. Is this a true statement:

> Anything requiring elevated privileges has to run outside of the python runner, and should use the `run-local` or `run-remote` runners with `sudo` enabled.

I would hate to get into the business of having to manage an elevated privilege worker at any point. It may come up in a fringe case where we would allow a user to adjust the privileges of specific workers, but easily :footprints: :arrow_right: :football:-able until later.

Kami:
@jfryman So for now you are OK with dropping flags which require root permissions and only scan TCP ports?

jfryman:
@Kami Yes. :smile:

Aamir-raza-1:
@Kami , @jfryman .. I will drop the flags that require root permissions and commit the code.

Kami:
I've pushed a change so now the action doesn't require root priveleges - e74ffa4ddb031efc863a4dde0a1c8c4ae4cdf517

I also pushed some other changes and fixes:

1. Lint and style fixes - af9daf9420ef9693d3a9081315f7fc2a9e30594a
2. Changed the parameter name from `host` to `ip_address` to be more explicit - 566ad74c2306b8e36bbc5e98021cb96b716ccaec. Passing a hostname to the `scan` method doesn't work correctly. Exception which is shown bellow is returned:

```
  File "/data/st2contrib/packs/linux/actions/get_open_ports.py", line 41, in <module>
    PortScanner().run(host='stackstorm.com')
  File "/data/st2contrib/packs/linux/actions/get_open_ports.py", line 30, in run
    'state': ps[host][comm][port]['state'],
  File "/data/stanley/virtualenv/local/lib/python2.7/site-packages/nmap/nmap.py", line 567, in __getitem__
    return self._scan_result['scan'][host]
KeyError: 'stackstorm.com'
```

Kami:
@Aamir-raza-1 Sorry, hope I didn't step on your on your toes. I just noticed the comment after I pushed the changes.

PR #100, opened 2015-01-25
live hosts

Aamir-raza-1:
This pack scans live hosts on a given network using nmap. It returns total number of live hosts and a list of IP's of the live hosts on a network. It takes one argument that is the network, which is mandatory
Usage:
st2 run get_live_hosts.get_live_hosts network=<1.1.1.1/1> # give the network(a subnet) for listing the live hosts

### DIFF
--- a/packs/get_live_hosts/actions/get_live_hosts.py
+++ b/packs/get_live_hosts/actions/get_live_hosts.py
@@ -1,0 +1,26 @@
+import nmap
+from st2actions.runners.pythonrunner import Action
+
+
+
+class check_live_hosts(Action):  
+
+  def run(self, network):
+    host_list = []
+    res = nmap.PortScanner()
+    print "Starting Network Discovery...."
+    lh = res.scan(network, arguments='-sP -n')
+    print "Number of Live Hosts on network : ",lh['nmap']['scanstats']['uphosts']
+    print "Live Hosts : "
+    for ip in res.all_hosts():
+      host_list.append(ip)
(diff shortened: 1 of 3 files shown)


packs/get_live_hosts/actions/get_live_hosts.py
```diff
@@ -0,0 +1,26 @@
+import nmap
+from st2actions.runners.pythonrunner import Action
+
+
+
+class check_live_hosts(Action):
```
Kami: Minor style thing - please use camel case for class name (`CheckLiveHostsAction`).

Kami:
Same here - this action also fits better in the existing linux pack.

packs/linux/actions/get_live_hosts.yaml
```diff
@@ -0,0 +1,11 @@
+name: get_live_hosts
+runner_type: run-python
+description: List all hosts on a given Network
+enabled: true
+entry_point: get_live_hosts.py
+parameters:
+  network:
+    type: string
+    description: "Network to query live hosts "
```
Kami: Would be more explicit to say `Subnet of a network to scan for live hosts (e.g. 192.168.0.0/24)`. Probably also better to call the parameter `subnet`.

Aamir-raza-1: yeah, You are right, I have changed the name from network to subnet and remove the main entry point.

PR #104, opened 2015-01-27
chef integration pack initial commit

dennybaa:


dennybaa:
Sorry, guys! I messed up with the pull request. Ignore it so far.
Thank you.

packs/chef/actions/lib/shell_out.py
```diff
@@ -0,0 +1,23 @@
...
+def shell_out(command, env={}):
+    env = os.environ.copy().update(env)
+    proc = subprocess.Popen(args=command, stdin=None, stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE, env=env)
+    while True:
+        fdlist = [proc.stdout.fileno(), proc.stderr.fileno()]
+        ios = select.select(fdlist, [], [])
+        for fd in ios[0]:
+            if fd == proc.stdout.fileno():
+                sys.stdout.write(proc.stdout.read())
```
Kami: I believe .read() blocks if there is no data available, right?

dennybaa: Thanks Kami for pointing out here, I wanted to use readline() it's a typo... This code is not so nice, maybe you might to know how to check how many bytes are available in the stream. It will be much better than readline() I suppose.

Kami: Yeah, I think readline should work.

Also, it would be nice to make this function a bit more generic (allow user to pass in streams / file-like objects to which the stdout and stderr will be written) and move it to https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/shell.py

dennybaa: Tomaz, readline() is a bad idea also as it happened. http://gyazo.com/6a954c05aa2d75b72c799205028b5d31
read(io.DEFAULT_BUFFER_SIZE) doesn't harass CPU and it's about ~ 9%. read(n) call is non-blocking it well return less bytes if n available.
About common utils it's cool, agreed. But what if it's a remote script runner on a remote host without st2 pre-installed?

Kami: That's a good point - I just quickly glanced over it and I didn't notice it's a remote action. Using st2common won't work with remote actions if st2common is not installed there so yeah, bundling this function inside the pack is fine.

PR #107, opened 2015-01-28
iostat action

Aamir-raza-1:
Iostat action returns I/O stats of system storage.

### DIFF
--- a/packs/linux/actions/iostat.yaml
+++ b/packs/linux/actions/iostat.yaml
@@ -1,13 +1,24 @@
 name: "iostat"
 runner_type: "run-remote"
-description: "returns I/O stats of system storage"
+description: "reutrns I/O stats of system storage"
 enabled: true
 entry_point: ""
 parameters:
+  count:
+    description: "repeat the display count times"
+    default: 1
+  cpu:
+    type: 'boolean'
+    description: "Display CPU"
+    default: true
+  device:
+    description: "display only device stats"
+    type: 'boolean'
+    default: false
   cmd:
     immutable: true
     default: "iostat {{args}}"
     position: 0
   args:
     description: "Command line arguments"
-    default: " " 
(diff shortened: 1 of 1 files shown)


jfryman:
This is a good start. Would it be possible to also add some of the flags used with `iostat`?

``` 
    -C    Display CPU statistics.  This is on by default, unless -d is specified.

     -c    Repeat the display count times.  If no wait interval is specified, the default is 1 second.

     -d    Display only device statistics.  If this flag is turned on, only device statistics will be displayed, unless -C or -U or -T is also specfied to enable the display of CPU, load average or TTY statistics.

     -I    Display total statstics for a given time period, rather than average statistics for each second during that time period.

     -K    In the blocks transferred display (-o), display block count in kilobytes rather then the device native block size.

     -n    Display up to devs number of devices.  iostat will display fewer devices if there aren't devs devices present.

     -o    Display old-style iostat device statistics.  Sectors per second, transfers per second, and miliseconds per seek are displayed.  If -I is specified, total blocks/sectors, total transfers, and miliseconds per seek are displayed.

     -T    Display TTY statistics.  This is on by default, unless -d is specified.

     -U    Display system load averages.  This is on by default, unless -d is specified.

     -w    Pause wait seconds between each display.  If no repeat count is specified, the default is infinity.
```

Aamir-raza-1:
@jfryman , Thanks for your valuable input. Yeah sure, I will do this shortly and push it up the stream.

packs/linux/actions/iostat.yaml
```diff
@@ -0,0 +1,24 @@
+name: "iostat"
+runner_type: "run-remote"
+description: "returns I/O stats of system storage"
+enabled: true
+entry_point: ""
+parameters:
+  count:
```
Kami: Looks like this parameter is not used in args.

Besides that I tested the action and it looks good to me.

PR #110, opened 2015-01-29
chefpack: omnibus chef installation method

dennybaa:


packs/chef/actions/lib/shellhelpers.py
```diff
@@ -0,0 +1,62 @@
+import argparse
+import sys
+import os
+import subprocess
+import select
+import io
+
+# Starts command in subprocess processing output as it appears.
+def shell_out(command, env={}):
```
Kami: Defaulting argument value to a mutable object (list, dictionary) is usually a bad idea and will result in unexpected behavior - https://stackoverflow.com/questions/1132941/least-astonishment-in-python-the-mutable-default-argument

Everywhere you default an argument to an empty dict or list, you should default the argument to `None` and inside the function do `value = value or {}` or `value = value or []`.

packs/chef/actions/lib/shellhelpers.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def long_arglist(self, kwargs={}):
+        _arg_list(kwargs or self.parse(), short=False)
+
+    def _arg_list(self, kwargs, short=None):
+        '''
+        Returns list of command line arguments with short or long key names.
+        '''
+        cmd = []
+        for n, v in (self.args or self.parse()).items():
```
Kami: Looks like `self.args` is not defined anywhere.

Did you intend to cache parsed args in `self.args` or?

jfryman:
A few changes to be made. Going to merge so I can edit.

packs/chef/actions/lib/shellhelpers.py
dennybaa: Thanks Thomaz for explanaition, it's  an amazing feature :) I couldn't just imagine it that it works kinda static variables...

packs/chef/actions/lib/shellhelpers.py
dennybaa: I haven't been thinking about caching it though. In case, say it, I want to run short_arglist several time with different kwargs. But it's had no use finally...

packs/chef/actions/lib/shellhelpers.py
dennybaa: Does this look sane now https://github.com/dennybaa/st2contrib/blob/chef-pack/packs/chef/actions/lib/shellhelpers.py ?

PR #113, opened 2015-02-01
Fix newrelicsensor handler

manasdk:
* Simplified things to get around some flask routing.
* Somewhat straight forward responses.
* Added a bit of logging to improve visiblity.

### DIFF
--- a/packs/newrelic/sensors/newrelic_application_sensor.py
+++ b/packs/newrelic/sensors/newrelic_application_sensor.py
@@ -17,8 +17,7 @@ import eventlet
 import requests
 import sys
 
-from flask import jsonify, request, Flask
-from six.moves import http_client
+from flask import request, Flask
 from six.moves import urllib_parse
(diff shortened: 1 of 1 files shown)


packs/newrelic/sensors/newrelic_application_sensor.py
```diff
@@ -240,6 +239,10 @@ def _dispatch_server_normal(self, payload, attempt_no=0):
             eventlet.spawn_after(self._normal_report_delay, self._dispatch_server_normal,
                                  payload, attempt_no + 1)
 
+    def _dispatch_trigger(self, trigger, payload):
+        self._sensor_service.dispatch(trigger, payload)
+        self._log.info('Dispatched %s with payload %s.', trigger, payload)
```
Kami: Hm, I thought `self._sensor_service.dispatch` already logs it, but it looks like it doesn't. It would be a good idea to also log it there under debug log level.

PR #114, opened 2015-02-01
Chef pack cleanup

Kami:
/cc @dennybaa

### DIFF
--- a/packs/chef/actions/lib/omnibus.py
+++ b/packs/chef/actions/lib/omnibus.py
@@ -1,40 +1,41 @@
 import shellhelpers as shell
 
+
 class Omnibus(object):
     OMNITRUCK = 'https://www.chef.io/chef/install.sh'
 
     def __init__(self, options):
         supports = ('pre_release', 'version', 'download_path')
-        self.options = { k: options[k] for k in supports }
+        self.options = {k: options[k] for k in supports}
 
-    def build_command(self, version=None, pre_release=False, download_path=None):
+    def build_command(self, version=None, pre_release=False,
+                      download_path=None):
         command = "curl -sL %s | sudo bash" % self.OMNITRUCK
 
-        if pre_release or version != None or download_path != None:
-          command += " -s --"
+        if pre_release or version or download_path:
+            command += " -s --"
 
         if pre_release:
-          command += " -p"
+            command += " -p"
 
-        if version != None:
-          command += " -v \"%s\"" % version
+        if version:
+            command += " -v \"%s\"" % version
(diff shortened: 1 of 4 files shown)


packs/chef/actions/lib/omnibus.py
```diff
@@ -1,40 +1,51 @@
...
+        command_args = [pipes.quote(arg) for arg in command_args]
+        command_args = ' '.join(command_args)
+        command = command + ' ' + command_args
 
+        return command
 
     def chef_installed(self):
         knife = "/opt/chef/bin/knife"
-        command = "test -x %s" % knife
+        command = "test -x %s" % (pipes.quote(knife))
```
dennybaa: Why do you suggest using pipes.quote? It's not required in the case above.
Also pipes.quote is a deprecated function (https://docs.python.org/2/library/pipes.html#pipes.quote), so it's better not to use  since it makes code not forward compatible.

Kami: Yeah, as it stands right now, escaping is not required, but in cases like this (when dealing with security and possible injections) it's way better to be safe than sorry.

Let's say that someone makes path to knife binary user configurable via parameter and doesn't add / forgets to add pipe.quote - bam, here we have possible shell injection.

And if you use `subprocess.Popen` with a args list and not a command string, Popen will do the same for you internally (automatically escape every argument).

As far as backward compatibility goes - `pipes.quote` is still available in Python 3.3 and 3.4 so we should be fine for  a long time if you consider how slow Python 3 adoption is and how long before a new major release where this function might be removed comes out.

Kami: And because of that, when possible, you should aim to use args list instead of string.

But yeah, that's not always possible in an easy manner, especially when you are dealing with commands which contain pipes.

dennybaa: I agree about security and thank you for the point. Also cool to know that it will work in Python 3.3.
But we should keep in mind also  what is crucial to escape and what is not. Regarding that the path configurable or not, the user has already remote access via ssh isn't he? And even more full sudo privileges, thus consequently the "securing" code doesn't make the whole solution anyhow more secure...

Kami: That's true right now, but in the future we will add support for RBAC and then a user might only be granted access to run a particular action or set of actions - then escaping will be important.

dennybaa: Okay than. I will keep in mind command line escaping.

PR #117, opened 2015-02-01
Slack Sensor Fixes

jfryman:
A few things to change to prevent the sensor from dying randomly.

* Adds  `timestamp` schema
* Adds bypass for `message_updated` event.
* Fix startup username/channel caching

/cc @Kami

### DIFF
--- a/packs/slack/sensors/slack_sensor.py
+++ b/packs/slack/sensors/slack_sensor.py
@@ -1,6 +1,7 @@
 import json
 
 import eventlet
+import time
 from slackclient import SlackClient
 
 from st2reactor.sensor.base import PollingSensor
@@ -21,7 +22,7 @@ class SlackSensor(PollingSensor):
         self._logger = self._sensor_service.get_logger(__name__)
         self._token = self._config['sensor']['token']
         self._handlers = {
-            'message': self._handle_message
+            'message': self._handle_message,
         }
 
         self._user_info_cache = {}
@@ -35,7 +36,8 @@ class SlackSensor(PollingSensor):
             msg = 'Failed to connect to the Slack API. Invalid token?'
             raise Exception(msg)
 
-        self._populate_cache(user_data=data['users'], channel_data=data['channels'])
+        self._populate_cache(user_data=self._api_call('users.list'),
+                                channel_data=self._api_call('channels.list'))
 
     def poll(self):
         result = self._client.rtm_read()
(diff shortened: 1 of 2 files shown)


packs/slack/sensors/slack_sensor.py
```diff
@@ -75,29 +78,35 @@ def _handle_result(self, result):
...
-            'channel': {
-                'id': channel_info['id'],
-                'name': channel_info['name'],
-                'topic': channel_info['topic']['value'],
-            },
-            'timestamp': int(float(data['ts'])),
-            'text': data['text']
-        }
-        self._sensor_service.dispatch(trigger=trigger, payload=payload)
+        if data.has_key('subtype'):
```
jfryman: Is there a more Pythonic way to express this?

packs/slack/sensors/slack_sensor.py
```diff
@@ -35,7 +36,8 @@ def setup(self):
             msg = 'Failed to connect to the Slack API. Invalid token?'
             raise Exception(msg)
 
-        self._populate_cache(user_data=data['users'], channel_data=data['channels'])
+        self._populate_cache(user_data=self._api_call('users.list'),
```
Kami: Hm, the whole idea was to avoid unnecessary API calls and re-use information why is already returned on `rtm.start`.

Did the API change or is it something else?

packs/slack/sensors/slack_sensor.py
```diff
@@ -75,29 +78,35 @@ def _handle_result(self, result):
     def _handle_message(self, data):
         trigger = 'slack.message'
```
Kami: To avoid nesting and make it a bit more readable, I would just do:

```python
if 'subtype' in data:
    # Ignore events for edited messages
    return

user_info = ...
```

packs/slack/sensors/slack_sensor.py
jfryman: The API must have changed. These attributes are no longer available. But, this seeds the cache and they are resolved with https://github.com/StackStorm/st2contrib/blob/slack-updates/packs/slack/sensors/slack_sensor.py#L111-L123 and put into cache.

Kami: I see and yeah, I think some of the recent API changes were also the reason for `is_admin` stuff breaking.

And yeah, as long as we seed it with all the users at the beginning I'm fine with it.

jfryman:
:rage: Sorry. Wrong button. I'll make any fixes in a subsequent PR.

PR #119, opened 2015-02-01
chef pack ohai uses remote script runner

dennybaa:


packs/chef/actions/chef_install.py
```diff
@@ -33,15 +33,15 @@ def install(self, install_method, options):
     def execute(self):
         self.parser = shell.CmdlineParser(self.cmdline_options)
         options = self.parser.parse()
-        install_method = [i for i in options.keys() if i in self.SUPPORTED_METHODS ]
+        install_method = options['method']
 
-        if len(install_method) > 1:
-            sys.stderr.write("You can't use more than one installation method!\n")
-            sys.stderr.write("Choose one of the following methods: %s\n" % ', '.join(self.SUPPORTED_METHODS))
+        if not (install_method in self.SUPPORTED_METHODS):
```
Kami: A bit more pythnoic would be to say `if install_method not in self.SUPPORTED_METHODS`

packs/chef/actions/lib/shellhelpers.py
```diff
@@ -43,17 +42,18 @@ def __init__(self, parser_options):
             self.parser.add_argument(k, lg, **kwargs)
             self._keyname[lg.lstrip('-')] = k
 
+
     def parse(self, argv=None):
-        argv = argv or sys.argv
```
Kami: I made this change because I prefer to be explicit - `parse_args` does indeed use `sys.arg`v if you don't explicitly provide `args` argument, but it's more explicit and clear if you explicitly pass it to the method.

packs/chef/actions/chef_install.py
dennybaa: Okay, cool. Is everything else ok to merge it?

packs/chef/actions/lib/shellhelpers.py
```diff
@@ -43,17 +42,18 @@ def __init__(self, parser_options):
...
     def short_arglist(self, kwargs=None):
-        kwargs = kwargs or {}
-        return self._arg_list(self.parse(), short=True)
+        return self._arg_list(kwargs or self.parse(), short=True)
+
 
     def long_arglist(self, kwargs=None):
-        kwargs = kwargs or {}
-        return self._arg_list(self.parse(), short=False)
+        return self._arg_list(kwargs or self.parse(), short=False)
```
Kami: Looks like kwargs is unused in `_arg_list` method.

dennybaa: That's a generic function, maybe it will be reused. It doesn't matter anyway. If someone including me wants to build argument list using provided argument dict he is free to it.

packs/chef/actions/lib/shellhelpers.py
Kami: Besides that, LGTM

dennybaa: Yeah but actually its parse_args behavior is implicit:) Which for `args=None` takes `sys.argv[1:0]` and if it's given it takes the whole list. So your changes will break my code:
```
       argv = argv or sys.argv
       return vars(self.parser.parse_args(args=argv)) # since I expect args without the first one this won't work
```

Kami: Ah, ok. In this case we should then default to `sys.argv[1:]` :)

dennybaa: So let it be just as it is)

dennybaa: Also it's a bit ironic. That devs of python stdlib who were also truly inspired with "explicit better than implicit paradigm" made such intricate behavior for parse_args )))

packs/chef/actions/chef_install.py
Kami: Yeah, go ahead.

PR #123, opened 2015-02-04
pager duty action

Aamir-raza-1:
Contians actions for 
1. launching an incident
2. Sending acknowledgment for launched incidents
3. Resolving Incidents
4. Retreiving all the open incidients

### DIFF
--- a/packs/pagerduty/actions/ack_incident.py
+++ b/packs/pagerduty/actions/ack_incident.py
@@ -1,0 +1,16 @@
+import json
+import requests
+from lib.action import PagerDutyAction
+
+
+class AckIncident(PagerDutyAction):
+  def run(self, event_key):
+    """Acknowledgment of a trigger, You can provide a list of keys"""    
+#    ack_keys = event_key.split(",")
+    ack_keys = [x.strip() for x in event_key.split(',')]
+   
+    for arg in ack_keys:
+      self.pager.acknowledge_incident(self.config['service_api'], arg)
+ 
+    return ack_keys
+
--- a/packs/pagerduty/actions/ack_incident.py
+++ b/packs/pagerduty/actions/ack_incident.py
@@ -2,12 +2,9 @@ from lib.action import PagerDutyAction
 
 
 class AckIncident(PagerDutyAction):
-  def run(self, event_key):
-    """Acknowledgment of a trigger, You can provide a list of keys"""    
-#    ack_keys = event_key.split(",")
-    ack_keys = [x.strip() for x in event_key.split(',')]
-   
-    for arg in ack_keys:
+  def run(self, event_keys):
+    """Acknowledgment of a trigger, You can provide comma(,) separated keys for acknowledging more events at once"""    
+    for arg in event_keys:
       self.pager.acknowledge_incident(self.config['service_api'], arg)
  
     return ack_keys
--- a/packs/pagerduty/actions/ack_incident.py
+++ b/packs/pagerduty/actions/ack_incident.py
@@ -1,5 +1,3 @@
-import json
-import requests
 from lib.action import PagerDutyAction
 
 
--- a/packs/pagerduty/actions/ack_incident.yaml
+++ b/packs/pagerduty/actions/ack_incident.yaml
@@ -4,7 +4,7 @@ description: ACK an incident on PagerDuty
 enabled: true
 entry_point: ack_incident.py
 parameters:
-  event_key:
-    type: string
+  event_keys:
+    type: array
     description: Key for event to be acknowledged, You can also Provide comma seprated keys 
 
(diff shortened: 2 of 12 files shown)


Kami:
Thanks, I will review it shortly.

Kami:
Looks like `requirements.txt` file is missing `pygerduty` dependency.

packs/pagerduty/actions/ack_incident.py
```diff
@@ -0,0 +1,16 @@
+import json
```
Kami: This import seems to be unused. Same goes with requests.

(applies to all python files)

packs/pagerduty/actions/ack_incident.yaml
```diff
@@ -0,0 +1,10 @@
+name: ack_incident
+runner_type: run-python
+description: ACK an incident on PagerDuty
+enabled: true
+entry_point: ack_incident.py
+parameters:
+  event_key:
+    type: string
```
Kami: You can use type `list`. This way the command line client will do the splitting for you.

Please also change the key name to ``event_keys`` so it's explicit and implies it can contain more than one key.

packs/pagerduty/actions/lib/action.py
```diff
@@ -0,0 +1,31 @@
...
+  
+  def _init_client(self):
+    api_key = self.config['api_key']
+    service_api = self.config['service_api']
+    subdomain = self.config['subdomain']
+    pager = pygerduty.PagerDuty(subdomain, api_token=api_key)
+    return pager
+
+##get all the acknowledged incidents
+  def ack_incidents(self):
```
Kami: To make it more explicit that this and the method bellow returns stuff, please add `get_` prefix (e.g. `get_ack_incidents`).

packs/pagerduty/actions/open_incidents.yaml
```diff
@@ -0,0 +1,6 @@
+name: open_incidents
```
Kami: I think it would be better to name the action `get_open_incidents` - this way it's clear it's retrieving and not opening a new incident.

Aamir-raza-1:
Ok sure, thanks for the suggestions, I will make the following changes and push up the steam

packs/pagerduty/actions/ack_incident.yaml
Aamir-raza-1: adding type list results in error.
u'list' is not valid under any of the given schemas
Is there any other keyword for type list?

Kami: Ugh, sorry, it's `array` :)

packs/pagerduty/config.yaml
```diff
@@ -0,0 +1,4 @@
+subdomain: 'devarchs'
```
jfryman: probably want to remove these before merge. :)

packs/pagerduty/actions/resolve_incident.yaml
```diff
@@ -0,0 +1,10 @@
+name: resolve_incident
+runner_type: run-python
+description: Resolve an incident whose key is provided
+enabled: true
+entry_point: resolve_incident.py
+parameters:
+  event_keys:
+    type: array
```
jfryman: Does this work as an array? If I supply a comma separated list, I think I'd run into a schema problem, no?

packs/pagerduty/actions/launch_incident.py
```diff
@@ -0,0 +1,9 @@
+from lib.action import PagerDutyAction
+
+
+class LaunchIncident(PagerDutyAction):
+  def run(self, description, details):
```
jfryman: Should `details=None` since it's not a required attribute?

packs/pagerduty/actions/ack_incident.yaml
```diff
@@ -0,0 +1,10 @@
+name: ack_incident
+runner_type: run-python
+description: ACK an incident on PagerDuty
+enabled: true
+entry_point: ack_incident.py
+parameters:
+  event_keys:
+    type: array
+    description: Key for event to be acknowledged, You can also Provide comma separated keys
```
jfryman: Same question here re: comma separated and array type

packs/pagerduty/actions/resolve_incident.yaml
Aamir-raza-1: I tested it by supplying comma seperated keys, and it runs fine. 
like for instance : 
 st2 run pagerduty.ack_incident event_key=21ab8308116,6ff46b1abbe9ee82965e764

packs/pagerduty/actions/launch_incident.py
Aamir-raza-1: Ah, yeah. You are right here. Let me push it after changing it

packs/pagerduty/actions/resolve_incident.yaml
Kami: Yeah, using the CLI you specify an array as a comma delimited string (inside the CLI string gets split into comma and converted into a list).

Will also add clarification on how to specify parameters which are a type of array to the CLI docs.

jfryman: 👍

PR #124, opened 2015-02-05
[jmx] Invoke method action

Kami:
This pull request adds `invoke_method` action to jmx pack which allows user to invoke arbitrary MBean method exposed via JMX.


manasdk:
Checking in a jar! We should `mvn install` via a bootstrap process.

(Ok. I am trolling here but in general we may at some point want to look at bootstrap for environments apart from python)

manasdk:
+1

Kami:
@manasdk As soon as we pull maven in, the whole universe will implode :P



Also, both of the jar dependencies we bundle here are old and not available in the maven repositories so in this case it wouldn't help much. But yeah, it might if we ever support Java as a first-class citizen, then it might if developers like to pull in different utility libraries.

PR #128, opened 2015-02-11
[aws] New service event notification sensor

Kami:
This pull request adds a new generic service notification to the AWS pack. This sensor exposes a HTTP interface (same as the new relic sensor) which listens for notification events from AWS services such as S3 which are sent via AWS SNS using HTTP(s) endpoint.

I wanted to avoid writing a custom sensor and just use webhook with a custom trigger, but sadly this doesn't work here.

The problem is that webhook payload body is JSON, but it contains an attribute `Message` which value is serialized version of JSON (string). This means we can't do any criteria matching or parameter capturing based on the values inside this attribute. And that's the attribute we are actually interested in since other attributes are just wrapper attributes for AWS SNS notification.

See https://gist.github.com/Kami/1834a822839b264e9897 for example payload.


packs/aws/README.md
```diff
@@ -1,32 +1,103 @@
...
+
+### ServiceNotificationsSensor
+
+This sensor exposes a HTTP interface and listens for service event
+notifications which are delivered via AWS SNS service using HTTP(s) endpoint.
+
+Currently it supports event notifications generated by the S3 service -
+http://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html
+
+Keep in mind that this sensor doesn't implement any kind of authentication.
```
Kami: We currently don't do that so we should also add a similar note to the new relic sensor - /cc @manasdk

manasdk: Yes good point. How would we support AUTH - at least with NR webhook I did not see many options.

I believe @DoriftoShoes does something with the github hook sensor running in ci/cd pipeline.

DoriftoShoes:
GitHub has a secret token that you set up for the webhook.  I put that token in the config file.  I also validate the GitHub ip address via an API they offer for that.

https://github.com/StackStorm/st2sandbox/blob/master/packs/webhooks/sensors/webhook_sensor.py#L22

None of this would probably work for these things though.  I'd have to research SNS and see if we could set up any sort of auth.

PR #129, opened 2015-02-11
Added newpack.sh

jtopjian:
This commit adds newpack.sh, a helper script to create a skeleton pack.


manasdk:
Joe,

Have you looked at https://github.com/StackStorm/st2sdk? Seems like we should add pointers from st2contrib to that tool.

Maybe that would work for you as well?

jtopjian:
I was not aware of that repo. Indeed that works for me :smile:

manasdk:
K. Will updates docs here to point to that tool.

PR #146, opened 2015-03-01
Softlayer basic pack

Itxaka:
Softlayer pack with libcloud as backend that allows for create/destroy instance and create/destroy keypair. Tested and working.

I will update this pack in the following weeks with more actions.


Kami:
Thanks, I will have a look shortly :)

packs/softlayer/actions/lib/softlayer.py
```diff
@@ -0,0 +1,36 @@
...
+        driver = self._get_driver()
+        nodes = [node for node in driver.list_nodes() if node.extra['hostname'] == kwargs['name']]
+        if len(nodes) == 1:
+            self.logger.info("Destroying node {} with remote ID {}".format(kwargs['name'], nodes[0].id))
+            return driver.destroy_node(nodes[0])
+        else:
+            self.logger.error("Node with name {} not found".format(kwargs['name']))
+            exit(1)
+
+    def _run_action(self, **kwargs):
```
Kami: Please explicitly declare all the arguments here - it makes it easier to see what's going on.

packs/softlayer/actions/create_instance.yaml
```diff
@@ -0,0 +1,41 @@
...
+enabled: true
+entry_point: run.py
+name: create_instance
+parameters:
+  action:
+    default: create_instance
+    immutable: true
+    type: string
+  name:
+    required: true
```
Kami: Please add a description for all the non-immutable parameters.

Kami:
Added some comments - there are some minor improvements and cleanups which need to be done. Besides that, looks good to me.

packs/softlayer/actions/delete_key_pair.yaml
```diff
@@ -0,0 +1,13 @@
+description: 'Deletes a keypair by name. If there are mutiple keys with the same name it will only delete the first'
+enabled: true
+entry_point: run.py
+name: delete_keypair
+parameters:
+  action:
+    default: delete_key_pair
+    immutable: true
+    type: string
+  key_pair:
```
Kami: Looks like this parameter needs to gold a key pair object name, right?

Please clarify that in the parameter description.

packs/softlayer/actions/create_key_pair.yaml
```diff
@@ -0,0 +1,13 @@
+description: 'Creates a keypair by name'
+enabled: true
+entry_point: run.py
+name: create_keypair
+parameters:
+  action:
+    default: create_key_pair
+    immutable: true
+    type: string
+  name:
```
Kami: Hm, looking at the code, the value of this argument is supposed to be a node hostname?

Or is it supposed to be a node name? or node.name == hostname?

packs/softlayer/actions/lib/softlayer.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _destroy_instance(self, **kwargs):
+        driver = self._get_driver()
+        nodes = [node for node in driver.list_nodes() if node.extra['hostname'] == kwargs['name']]
+        if len(nodes) == 1:
+            self.logger.info("Destroying node {} with remote ID {}".format(kwargs['name'], nodes[0].id))
+            return driver.destroy_node(nodes[0])
+        else:
+            self.logger.error("Node with name {} not found".format(kwargs['name']))
+            exit(1)
```
Kami: exit takes no arguments, please use `sys.exit`.

Itxaka:
@Kami Regarding the parameter names, Im just using the same ones that libcloud expects so I can pass kwargs across and have the functions work "magically" instead of having to check for the method name or have one python file per action.
This is something that was not really clear from st2 documentation. Is there any kind of convention on how to do things? Is one abstract script preferred or one per action? Is explicitly preferred or code simplicity?

Kami:
@Itxaka In general, we prefer to be explicit and have multiple action files (one per action), but the way you have it is also fine in this case.

As far as using kwargs goes - the problem with that is that it's really hard to debug and for other users which are not familiar with Libcloud or don't want to go check Libcloud's code to see what's going on.

Explicitly defining arguments does mean a bit more (maintenance) work, but it's easier to follow and see what's going on.

Itxaka:
@Kami Perfect! I thought checking other packs that brevity was preferred but I rather have separate files so I have more control onver the actions. Will redo the patch soon!

Itxaka:
@Kami I updated the PR.
Basically I changed from the abstract "everything is done magically" to a more decent approach of one entrypoint per action, with proper and easy args and straighforward actions.

Looks much better this time IMHO, so please let me know any issues you can see with it. :+1: 

BTW, I still dont see a lot on the docs on how to react to exceptions or failures in the actions. Do we just let them raise whatever error? Do we try to catch them and return a message with the traceback?

Kami:
@Itxaka Thanks, I will have a look shortly.

As far as exceptions go - if they are fatal, in the Python runner, you can just let them propagate / bubble up. Any uncaught exception will cause action to fail (status = failed) and the exception will be reported to the user.

Itxaka:
@Kami Perfect regarding the errors. I was waiting for that to make the last commit as I wanted to return a  better error than an IndexErrorr :smile:

packs/softlayer/README.md
```diff
@@ -0,0 +1,17 @@
+# Softlayer Integration Pack
+
+Pack which allows integration with [Softlayer](https://www.softlayer.com/).
+
+## Configuration
+
+* ``username`` - Softlayer username.
+* ``api_key`` - Api key.
+* ``region`` - Not used yet, should be the default region you want to work with for instances.
```
Kami: Later on when you add support for other APIs it would make sense to split configuration into `compute` and `storage` section (similar as Libcloud pack does it).

packs/softlayer/actions/create_instance.yaml
```diff
@@ -0,0 +1,47 @@
...
+    type: string
+    description: Hostname for the instance.
+  domain:
+    required: false
+    type: string
+    description: domain to attach to the instance (defaults to example.com).
+  cpus:
+    required: false
+    type: integer
+    description: Cpus for this instance (defaults to 1).
```
Kami: Good practice it to also set the same default value in the action metadata file for all the parameters with default values.

This way this default value is also displayed when running `st2 action -h`.

Itxaka: Good point, updating!

Kami:
Merged into master. Thanks.

PR #150, opened 2015-03-03
Unformat domains in Message Data

jfryman:
Quick and dirty regex replace of message formatting given by Slack. When typing in a domain (stackstorm.com), slack formats this to `<http://stackstorm.com|stackstorm.com>`. Not sure where is the best way to filter this out, but for now here it is.

### DIFF
--- a/packs/slack/sensors/slack_sensor.py
+++ b/packs/slack/sensors/slack_sensor.py
@@ -25,6 +25,10 @@ class SlackSensor(PollingSensor):
                                           poll_interval=poll_interval)
         self._logger = self._sensor_service.get_logger(__name__)
         self._token = self._config['sensor']['token']
+        if 'strip_formatting' in self._config['sensor']:
+            self._strip_formatting = self._config['sensor']['strip_formatting']
+        else:
+            self._strip_formatting = False
 
         self._handlers = {
             'message': self._handle_message,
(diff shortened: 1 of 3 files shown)


Kami:
@jfryman Ideally we would make this an action parameter (e.g. `strip_tags` / `strip_formatting` and default it to False).

jfryman:
@Kami so, I did actually try that, but my python sucks. Let me throw what I had and you can point the error of my ways

jfryman:
@Kami What am I doing wrong?

I also totally swiped your parameter name. I was using `unformat_messages`, which was bleg.

jfryman:
Heh. Retyping it did the needful. Must have fat-fingered something somewhere. This works now. :)

Anything else?

packs/slack/sensors/slack_sensor.py
```diff
@@ -24,6 +25,11 @@ def __init__(self, sensor_service, config=None, poll_interval=None):
                                           poll_interval=poll_interval)
         self._logger = self._sensor_service.get_logger(__name__)
         self._token = self._config['sensor']['token']
+        if 'strip_formatting' in self._config['sensor']:
```
Kami: Minor style thing - `self._strip_formatting = self._config['sensor'].get('strip_formatting', False)` is a bit more Pythonic and short :)

Kami: Please also add documentation for this new config option to the readme.

Besides that, LGTM.

jfryman: Ooh! I like that shorter operation. I was looking for a ternary operator of sorts... but that will do nicely.

jcockhren:
nice!

PR #152, opened 2015-03-05
Adding additional actions to get individual VM info

jfryman:
Add additional actions to support the Autoscale case.

### DIFF
--- a/packs/rackspace/actions/get_vm_ips.py
+++ b/packs/rackspace/actions/get_vm_ips.py
@@ -1,0 +1,44 @@
+from lib.action import PyraxBaseAction
+from lib.formatters import to_server_dict
+
+__all__ = [
+    'GetVMIPsAction'
+]
+
+class GetVMIPsAction(PyraxBaseAction):
+    def run(self, count=None, region=None, metadata=None):
+        if region:
+            cs = self.pyrax.connect_to_cloudservers(region=region)
+        else:
+            cs = self.pyrax.cloudservers
+
+        servers = cs.list()
+
+        result = []
+        for server in servers:
+            item = to_server_dict(server=server)
+
+            if metadata:
+                include = self._metadata_intersection(server=item,
+                                                      metadata=metadata)
+
+                if not include:
+                    continue
+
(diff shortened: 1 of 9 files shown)


packs/rackspace/actions/get_vm_ips.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: get_vm_ips
+runner_type: run-python
+description: List all the available vms by IP. Optionally filter by metadata and count
```
Kami: Looking at the code - this action returns all the available public IP addresses for all or for a subset of VMs, right?

packs/rackspace/actions/get_vm_ips.py
```diff
@@ -0,0 +1,44 @@
...
+            item = to_server_dict(server=server)
+
+            if metadata:
+                include = self._metadata_intersection(server=item,
+                                                      metadata=metadata)
+
+                if not include:
+                    continue
+
+            result.append(item['public_ips'][1])
```
Kami: IIRC, it takes a while for a public IP to be assigned to a server after the create request so to be on the safe side, we should probably do something along the lines of:

```python

public_ips = item.get('public_ips', [])

if not public_ips:
  continue
```

If we don't do that, it's possible that this line will throw if someone just booted a VM...

Kami: And since it's a list maybe we can just append and return all the ips - result.extend(public_ips)? :)

packs/rackspace/actions/get_vm_ips.py
```diff
@@ -0,0 +1,44 @@
...
+
+        if count:
+            return result[0:count]
+        else:
+            return result
+
+    def _metadata_intersection(self, server, metadata):
```
Kami: Ideally, we would refactor this method on a base `BaseServersAction` or something along those lines :)

Kami:
I'm gonna merge this and clean it up a bit afterwards since I need the same functionality for my demo, heh :)

PR #153, opened 2015-03-05
Increased wait sleep from 3 seconds to 10 seconds to avoid race condition

DoriftoShoes:
If a workflow creates a vm and calls this action it needs to sleep briefly to avoid a race condition where the API does not return the instance information.


Kami:
It seems like a more robust solution would be to refactor the polling code a bit (handle instance doesn't exist and just poll again later), but this also works for me right now :)

PR #154, opened 2015-03-08
New mailgun pack

Kami:
It initially contains an action for sending an email via Mailgun HTTP API.


packs/mailgun/actions/send_email.py
```diff
@@ -0,0 +1,31 @@
...
+from st2actions.runners.pythonrunner import Action
+
+__all__ = [
+    'SendEmailAction'
+]
+
+SEND_EMAIL_API_URL = 'https://api.mailgun.net/v2/%(domain)s/messages'
+
+
+class SendEmailAction(Action):
```
jfryman: Questions for myself here... Let's say I have an integration with another Mail sender (Postmark, Amazon SES, whatever), and the pack action also uses the same class name. What happens? Should we care about namespacing here?

Kami: Class names are local to the pack and the file so explicit namespacing is not necessary - as long as the class is in a different file and / or pack, there should be no problems.

Same goes for sensors, although sensors are a bit different - they explicitly specify class name in the metadata file, where in the python-runner action case, we just pull out the first class from the file (I'd actually like to make this more explicit and consistent with the sensor, eventually...).

jfryman: :cool:. Thanks.

jfryman:
:+1:

packs/mailgun/pack.yaml
```diff
@@ -0,0 +1,10 @@
+---
+name : mailgun
+description : st2 content pack containing mailgun integrations
+keywords:
+  - email
+  - mail
+  - mailgun
+version : 0.1
+author : st2-dev
+email : info@stackstorm.com
```
lakshmi-kannan: I usually use support@stackstorm.com

packs/mailgun/actions/send_email.py
```diff
@@ -0,0 +1,31 @@
...
+            'subject': subject,
+            'text': text
+        }
+
+        if html:
+            data['html'] = html
+
+        api_url = SEND_EMAIL_API_URL % {'domain': domain}
+        response = requests.post(api_url, auth=('api', api_key), data=data)
+        return response.status_code == httplib.OK
```
lakshmi-kannan: I am not sure returning False fails the action. I think you might need to raise an exception. Can you double check this part?

Kami: Yeah, I don't think so either.

And a good point overall - we do follow this `return True / False` pattern in other cases as well, but I agree with you and think we should throw if an action fails so this is reported correctly up the chain.

Kami: Pushed a fix in d6c22437a47a6f34213dd386a9e40110c64a8a12.

PR #155, opened 2015-03-09
Changes to PagerDuty README

jfryman:
Few minor changes to the PD Readme.

### DIFF
--- a/packs/pagerduty/README.md
+++ b/packs/pagerduty/README.md
@@ -1,25 +1,31 @@
 # PagerDuty 
-This action enables the integration of PD into stackstorm. It is capable of performing the following actions
+
+This action enables the integration of PagerDuty into StackStorm. It is capable of performing the following actions:
+
 1. List all the open incidents on PD for a subdomain
-2. Send acknowledgment of any incident or you can even send ACK for multiple incidents using comma seprated values
-3. Close and open incident or you can even close multiple open incidents by providing comma sperated values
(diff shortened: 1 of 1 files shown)


packs/pagerduty/README.md
```diff
@@ -1,25 +1,31 @@
...
-3. Close and open incident or you can even close multiple open incidents by providing comma sperated values
+2. Send acknowledgment of any incident(s)
+3. Close and open incident(s)
 4. Launch an incident by giving its details and description
 
-# Confgiuration
+# Configuration
 
-You need the following parameters to be in the config.yaml file
 `subdomain: `'name of subdomain '
```
Kami: It would be nice if this used bullet points so it's consistent with other readmes.

jfryman: Do we have a canonical README that we can agree on? We're all over the place on this point...

Kami: Most of the readmes I've touched (that's probably at least 60% of the packs) follow the "canonical" format. Here is an example - https://github.com/StackStorm/st2contrib/blob/master/packs/slack/README.md

I also want to get this merged - https://github.com/StackStorm/st2contrib/pull/87, but first I need to update a couple of readmes which are left which still don't conform to this format.

jfryman: Yeah, #87 would be stellar to consistency. :+1:

PR #156, opened 2015-03-15
Windows pack

Kami:
First implementation includes `wmi_query` action.

### DIFF
--- a/packs/windows/README.md
+++ b/packs/windows/README.md
@@ -1,0 +1,52 @@
+# Windows Integration Pack
+
+Pack which allows integration with Windows systems.
+
+## Actions
+
+### WMIQueryAction
+
+Action which runs a provided WQL query on a specified host and returns the
+result.
+
+For this action to work, WMI client for Linux (wmic) needs to be installed on
+the host where this action is running. Packages for Ubuntu and other systems
+are available at [https://www.orvant.com/packages/](https://www.orvant.com/packages/).
+
+#### Parameters
+
+* ``host`` - Host of a Window machine to run the query on.
+* ``username`` - Account username (defaults to ``Administrator``).
+* ``password`` - Account password.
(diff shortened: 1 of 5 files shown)


Kami:
The same action / information would also be useful in the sensor context. There are more actions like that and we should think about it and try to come up with some convention / pattern for handling that.

A simple solution would be to also implement a sensor which emits this information, but this results in a duplicated code, etc.

manasdk:
Wouldn't it be better to have a native windows runner that access native win32 APIs?

Kami:
@manasdk Well, this uses WMI (https://msdn.microsoft.com/en-us/library/aa384642%28v=vs.85%29.aspx) which is a native Windows thing. Among other things, WMI can also be accessed remotely from a host which has a client (either Windows or Linux).

So in short - unless we need to support really old Windows server version, it doesn't matter since WMI appears to be the future of (remote) data retrieval and management.

dzimine:
This is really cool. 

As a follow up, think about making it a runner - not necessarily windows runner, but a "normal" runner. And what would it take to convert it to the powershell runner?

Kami:
@dzimine I added some comments about my findings to the JIRA ticket (STORM-973).

We could make it a runner and get it to execute console and powershell commands using `winexe` which works on Linux. The problem with winexe is that latest stable version has some bugs and we would need to use a version from master. This means we would need to build a version for distros we support which is something we could ideally avoid.

manasdk:
There is no way we get proper windows integration without having a runner that executes on windows. I see this as a windows management pack. Can we say a windows runner is different from this pack?

manasdk:
After understanding the intergation point I am positive about this pack. We should limit/define its scope.

Kami:
I will get this merged now, but eventually this action should be made obsolete in favor of https://github.com/StackStorm/st2/pull/1287

PR #158, opened 2015-03-21
[RFC] RSS pack with RSS sensor (first sensor which utilizes trigger with parameters)

Kami:
This pull request adds a new RSS sensor which monitors RSS / Atom feeds for new entries.

This is also a first sensor outside of StackStorm core (timer and webhook trigger) which utilizes trigger parameters (yay, achievement unlocked!). Being a first such sensor means the are some issues.

Right now it's hard to configure this sensor - only way to do it is to create a rule. Creating a rule to create a corresponding `TriggerDB` object makes some sense for built-in triggers such as webhook and timer, but it makes little sense for custom triggers utilized by a sensor. This also ties a bit into the thing we talked about yesterday - when the events flow into the system, there is no expectation of something happening. Requiring a user to create a rule to configure a trigger with parameters kinda breaks this promise... Another "problem" with rules is that those triggers with parameters are never deleted. This is not a big issue for webhooks and timers, but it's an issue for sensors.

To make custom sensors easier to configure we should add a CLI command which allows user to create `TriggerDB` objects for `TriggerTypeDB` with parameters. Sadly, to make things confusing we already have `st2 trigger <action>` set commands, but those commands actually operate on `TriggerTypeDB` and not `TriggerDB` objects (sad panda).

Any suggestions for what the CLI command name should be? The best I can come up with is `st2 trigger create-with-parameters`. An alternative would be to rename `trigger` to `trigger-type`, but when we would also need to do the same with sensors (sensor -> sensor-type).

While working and testing this sensor I used cURL to create / configure a `TriggerDB` object.

`feed_trigger.json`

``` json
{
    "pack": "rss",
    "name": "stackstorm.feed",
    "type": "rss.feed",
    "parameters": {
      "url": "http://stackstorm.com/feed/"
    }
}
```

``` bash
curl -X POST -H "Content-Type: application/json" "http://127.0.0.1:9101/v1/triggers" -d @feed_trigger.json
```
## Open questions / TODO
- How should the CLI command for creating triggers look like? See above for more info.
- We should update sensor wrapper to only call trigger handler methods (add_trigger, update_trigger, remove_trigger) for triggers with parameters. See https://github.com/StackStorm/st2contrib/pull/158#discussion_r26893132 for more info
- Update other sensors where it makes sense to utilize trigger parameters instead of a config (git, file watch, twitter, etc.). This should be a separate PR of course.

### DIFF
--- a/packs/rss/sensors/rss_sensor.py
+++ b/packs/rss/sensors/rss_sensor.py
@@ -46,6 +46,8 @@ class RSSSensor(PollingSensor):
     def poll(self):
         feed_urls = self._feed_urls.keys()
 
+        self._logger.info('Processing %s feeds' % (len(feed_urls)))
+
         for feed_url in feed_urls:
             self._logger.info('Processing feed: %s' % (feed_url))
             processed_entries = self._process_feed(feed_url=feed_url)
--- a/packs/rss/sensors/rss_sensor.py
+++ b/packs/rss/sensors/rss_sensor.py
@@ -24,7 +24,7 @@ class MLStripper(HTMLParser):
 
 
 class RSSSensor(PollingSensor):
-    DATASTORE_NAME_SUFFIX = 'last_published_at'
+    DATASTORE_NAME_SUFFIX = 'last_timestamp'
 
     def __init__(self, sensor_service, config=None, poll_interval=None):
         super(RSSSensor, self).__init__(sensor_service=sensor_service,
@@ -35,7 +35,7 @@ class RSSSensor(PollingSensor):
         self._logger = self._sensor_service.get_logger(__name__)
 
         config = self._config['sensor']
-        self._use_published_at_filtering = config.get('use_published_at_filtering', True)
+        self._use_timestamp_filtering = config.get('use_timestamp_filtering', True)
 
         # Stores a list of monitored feed urls
         self._feed_urls = {}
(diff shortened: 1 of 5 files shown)


packs/rss/sensors/rss_sensor.yaml
```diff
@@ -0,0 +1,28 @@
+---
+  class_name: "RSSSensor"
+  entry_point: "rss_sensor.py"
+  description: "Sensor which monitors RSS / Atom feed for new entries"
+  poll_interval: 15
+  trigger_types:
+    -
+      name: "feed"
```
Kami: Currently there is a another issue with triggers with parameters - we call `add_trigger` on sensor instance for every trigger which belongs to that sensor.

In this case this means we also call it for `rss.entry` triggers which makes little sense (this is an outbound trigger which is dispatched by a sensor and not something a sensor is interested in).

I propose to make a change to only call trigger handler methods for triggers with parameters.

Feedback?

Kami:
@dzimine @lakshmi-kannan @manasdk et al. This works as it is, but I want feedback and address open questions mentioned above.

packs/rss/sensors/rss_sensor.py
```diff
@@ -0,0 +1,223 @@
...
+from six.moves.html_parser import HTMLParser
+
+from st2reactor.sensor.base import PollingSensor
+
+__all__ = [
+    'RSSSensor'
+]
+
+
+class MLStripper(HTMLParser):
```
lakshmi-kannan: So you weren't joking :P

manasdk:
We have really messed up what should be a trigger with parameters and what should be config for a sensor.

manasdk:
@Kami https://gist.github.com/manasdk/6ffe8d8a4a2400b274dc

A continuation of the sentiment that we don't have the right model.

packs/rss/sensors/rss_sensor.py
Kami: lol.

I actually just noticed I fat fingered it and forgot "HT" prefix, lol.

Kami:
To additionally clarify my reasoning about trigger parameters - first I was for a new concept called "sensor instance" / "config template" where each sensor instance would get a different config.

This was before I truly understood triggers with parameters though. Now I'm against it because we already have an existing concept and solution for exactly this problem.

Only downside of this approach is that it could become a performance bottle neck if user wants to use a single sensor to monitor many different resources (e.g. 100's of jira instances / files / rss feeds).

This could come up in a SaaS scenario where a user would potentially use a single StackStorm instance. The thing is that there are many other problems around multi-tenancy right now so just "solving" it for sensors doesn't really help much.

As mentioned on Slack, I'm only "+1" for sensor instance (or similar) concept if we nuke triggers with parameters. I'm strongly against (and will veto any change) which introduces yet another concept without removing or simplifying existing concepts - complicated products sucks.

We already have two things and concepts now (sensor config for global sensor default values and trigger parameters) and additional concept is a no go - it adds yet another thing user needs to learn and keep in mind (increases learning curve), makes the whole thing a lot more complex and complicated, we now have two similar concepts (trigger with parameters and instances) - adds confusion, etc.

In short - it's only an improvement if the solution we go with solves the problem and doesn't introduce additional complexities or makes the product simpler and easier to use. If it solves the problem, but it doesn't meet the second criteria (aka increases the learning curve and introduces yet another concept) then it's just a bloat and something which should not go in the project.

manasdk:
> As mentioned on Slack, I'm only "+1" for sensor instance (or similar) concept if we nuke triggers with parameters. I'm strongly against (and will veto any change) which introduces yet another concept without removing or simplifying existing concepts - complicated products sucks.

We are still building understanding of our system. I've had strong opinions about this in the past about this thing too :)

I am willing to accept your position if we all agree it is the right model. Lets not talk about veto even before we have any discussion about the use-case, user expectation etc. :P

Kami:
Closing this since we never came to a consensus.

PR #159, opened 2015-03-21
victorops integration pack

Aamir-raza-1:
Victor ops integration pack allows to register an event and resolving it using stackstorm.
It requires two mandatory and one optional paramter
Mandatory parameters are:
1. severity: it can be CRITICAL, INFO, WARNING, ACKNOWLEDGMENT, RECOVERY
2. entity: entity is a id which victorps track for events and this is userd for acking or recovery of events
Optional parameter is message, message is the description of the event that you want to launch
Usage:
st2 run victorops.open_incident severity='CRITICAL' entity='Host' message='argumented'

If you dont want to give any message then
st2 run victorops.open_incident severity='RESOLVED' entity='Host''


packs/victorops/actions/open_incident.py
```diff
@@ -0,0 +1,17 @@
...
+class OpenIncident(VictorOpsAction):
+  def run(self, severity, entity, message=None):
+    prms = {
+    "message_type": severity,
+    "timestamp": int(time.time()),
+    "entity_id":entity,
+    "state_message":message
+    }
+    post_data = json.dumps(prms).encode()
+    data = urllib.urlopen(self.url, post_data)
```
Kami: Please use `request` package (`request.post`). There are multiple benefits over using a built-in urllib:

1. No issues with differences between Python 2.x and 3.x
2. request perform ssl cert validation

packs/victorops/actions/open_incident.py
```diff
@@ -0,0 +1,17 @@
+import urllib
+from lib.action import VictorOpsAction
+import json
+import time
+
+class OpenIncident(VictorOpsAction):
+  def run(self, severity, entity, message=None):
+    prms = {
+    "message_type": severity,
```
Kami: Please follow the pep8 style guide - https://www.python.org/dev/peps/pep-0008/

You can run pep8 tool on this file to view the pep8 style violations.

Kami:
Please add a readme which follows the same format as this one - https://github.com/StackStorm/st2contrib/blob/master/packs/slack/README.md

packs/victorops/actions/open_incident.py
Aamir-raza-1: Ok cool, sure thing

Aamir-raza-1: Done, Adding a Readme in short time

packs/victorops/pack.yaml
```diff
@@ -0,0 +1,5 @@
+name: Victorops
+description: Packs which allows integration with Victorops events.
```
Kami: Please also add keywords attribute.

See https://github.com/StackStorm/st2contrib/blob/master/packs/chef/pack.yaml#L4 for example.

packs/victorops/config.yaml
```diff
@@ -0,0 +1,4 @@
+api_key: 20131114/alert/8f048ed9-66af-4406-aa9c-35facef20829
```
jfryman: Probably need to remove this. :)

packs/victorops/config.yaml
```diff
@@ -0,0 +1,4 @@
...
+routing_key: even_team
+debug: false
+url: https://alert.victorops.com/integrations/generic
```
jfryman: Is there ever a situation where this would need to be changed?

jfryman:
Great start. A few things.

1. I am concerned with the namespace overloading. It does not seem like it will be easy to use this action, given that the action itself reads `open_incident` as the name, but you can actually open an incident, ack an incident, and close an incident. If I needed to do historical searching for actions, it would be very difficult to get history information as-is. I think we need to have individual actions for each of the state changes (open, ack, close) just like the PagerDuty pack.

2. I think there needs to be some discovery mechanism here too. How can I find/discover what entity_id I want to take action on without going to the website.

3. Needs a README to be in `st2contrib`

packs/victorops/config.yaml
Aamir-raza-1: Yes, actually this would be changed for every account who wants to use API, URL would be same but api key and routing key needs to be changed. I will add a readme for how to obtain these api key and routing key.

Aamir-raza-1:
Ok sure, we can detangle that, I wrote one generic funtion so that redundancy should be removed, but you have a valid point here. Let me work on detangling it and make sperate actions.

PR #162, opened 2015-03-23
adds polling sensor for rabbitmq queues

Itxaka:
adds a polling sensor to rabbitmq, that checks some queues and triggers the messages body and queue.

Provides: Polling sensor
Triggers: rabbitmq.new_message
Payload: dict(queue: string, body:object)


packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -0,0 +1,46 @@
+from st2reactor.sensor.base import PollingSensor
+import pika
```
lakshmi-kannan: Just curious, why did you settle on pika? We use kombu ourselves and debating a move to Pika. Would love to hear your thoughts.

manasdk:
Failure due to another pack. Someone should take a look.

packs/rabbitmq/sensors/queues_sensor.py
Itxaka: Basically because Its what I played around and what I know how to use ;)

Tried to use Kombu some time ago for a work project, but there were some stoppers that we were unable to get over. Moving it to pika took a couple of days and there was no issues on anything since then and we have added more and more stuff.

So looks like a personal choice more than technical reasons to be honest.

Kami:
Thanks, I will have a look shortly.

One thing which is worth noting inside the readme is that this sensor should only be used with `fanout` and `topic` exchanges - this way it doesn't affect the behavior of the app since messages will still be delivered to other consumers / subscribers.

If it's used with direct or headers exchanges, those messages won't be delivered to other consumers so it will affect app behavior and potentially break it.

Kami:
@manasdk Pushed a revert - that commit was pushed accidentally.

Syncing this branch with master should fix it.

Itxaka:
@Kami Good point, adding it to the readme

packs/rabbitmq/config.yaml
```diff
@@ -0,0 +1,6 @@
+sensor_config:
+  poll_time: 5
+  queues:
```
Kami: YAML also supports lists / arrays so you can use that instead of a comma delimited string.

For example:

```yaml
queues:
  - queue1
  - queue2
  -...
```

Itxaka: I feel that way translates pretty poorly to the readme in order to make it easier to know what it needs to be there.

Kami: I agree about this so feel free to add example to the readme and / or config.

Itxaka:
@Kami updated it to use a proper yaml list and clarified with an example on the readme.

Kami:
Thanks, LGTM.

I will check it out locally, test it and merge it after looks OK :)

packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -0,0 +1,45 @@
...
+        host = self._config['sensor_config']['host']
+        creds = PlainCredentials(username=self.user, password=self.password)
+        self.conn = pika.BlockingConnection(pika.ConnectionParameters(host=host, credentials=creds))
+        self.channel = self.conn.channel()
+        self.channel.basic_qos(prefetch_count=1)
+
+    def poll(self):
+        for queue in self.queues:
+            queue_state = self.channel.queue_declare(queue=queue, durable=True)
+            if queue_state.method.message_count != 0:
```
Kami: Another thing, shouldn't we do a while loop here and consume until message_count == 0?

Otherwise we will always only consume a maximum of one message every poll interval, even if there are multiple messages in the queue.

Or was this a conscious decision?

Kami:
Thanks.

I've encountered some minor issues during testing which I have fixed in a separate PR which is based on yours - https://github.com/StackStorm/st2contrib/pull/164.

I will close this PR and we can continue there.

packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -0,0 +1,45 @@
+from st2reactor.sensor.base import PollingSensor
+import pika
+from pika.credentials import PlainCredentials
+
+
+class RabbitMQSensor(PollingSensor):
```
lakshmi-kannan: For the next time, we should be more explicit with the class name. I am assuming we'll write more such sensors. RabbitMQSensor is way too generic :P.

Kami: I agree, I'm fine if you push rename to master :P

PR #163, opened 2015-03-24
Initial commit Nest pack

jfryman:
This pull request introduces the IoT pack Nest, allowing management of Nest devices via StackStorm.

## Supported Actions
```
+----------------------+------+-----------------+---------------------------------+
| ref                  | pack | name            | description                     |
+----------------------+------+-----------------+---------------------------------+
| nest.get_humidity    | nest | get_humidity    | Get the current humidity        |
| nest.get_mode        | nest | get_mode        | Manage nest modes               |
| nest.get_temperature | nest | get_temperature | Get the current temperature.    |
| nest.set_away        | nest | set_away        | Set nest to away mode           |
| nest.set_fan         | nest | set_fan         | Manage fan state                |
| nest.set_home        | nest | set_home        | Set nest to home mode           |
| nest.set_humidity    | nest | set_humidity    | Set humidity goal for nest      |
| nest.set_mode        | nest | set_mode        | Set current operating mode      |
| nest.set_temperature | nest | set_temperature | Set current temperature.        |
| nest.show            | nest | show            | Show current Nest information   |
| nest.toggle_away     | nest | toggle_away     | Toggle current Home/Away status |
+----------------------+------+-----------------+---------------------------------+
```


packs/nest/actions/fan.yaml
```diff
@@ -0,0 +1,24 @@
...
+runner_type: run-python
+description: Manage fan state
+enabled: true
+entry_point: fan.py
+parameters:
+  state:
+    type: string
+    description: State to set the fan
+    enum:
+      - "on"
```
Kami: Hm, there is no "off"? :)

packs/nest/actions/fan.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - auto
+    default: auto
+    required: true
+  structure:
+    type: integer
+    description: ID of Nest Home name
+    default: 0
+  device:
```
Kami: It might make sense to do what we do with some other actions - allow user to set global default for `device` and `structure` in the config, but also allow to override it on per action basis.

packs/nest/actions/humidity.py
```diff
@@ -0,0 +1,14 @@
+from lib import actions
+
+class HumidityAction(actions.BaseAction):
+    def run(self, structure, device, target=None):
+        if target:
+            target_humidity = target
+        else:
+            target_humidity = self._get_device(structure, device).humidity
```
Kami: Hm, does this case (not providing a target) even makes sense?

It looks like it just synchronizes all the devices to same target of the selected device?

If my observation is correct and this case still makes sense, we should document this in the parameter description.

packs/nest/actions/mode.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: mode
+runner_type: run-python
+description: Manage nest modes
+enabled: true
+entry_point: mode.py
+parameters:
+  mode:
+    type: string
+    description: State to set the fan
```
Kami: Need to update this :)

packs/nest/actions/temperature.yaml
```diff
@@ -0,0 +1,34 @@
...
+    description: Set target lower temperature for range
+  temp_high:
+    type: integer
+    description: Set target upper temperature for range
+  scale:
+    type: string
+    description: Scale to use when setting temperature
+    enum:
+      - fahrenheit
+      - celcius
```
Kami: `celcius` -> `celsius` :)

Or does nest actually relies on a misspelled word - http://search.cpan.org/~kedar/Device-Nest-0.09/lib/Device/Nest.pm?

packs/nest/actions/fan.yaml
jfryman: Weird, right? It's either `on`, or `We know best` mode.

packs/nest/actions/temperature.yaml
jfryman: No, I don't get to used a civilized temperature scale often, so I often misspell. I'll fix. :)

packs/nest/actions/mode.yaml
jfryman: :+1:

packs/nest/actions/humidity.py
jfryman: So, I actually debated a pattern here in my head, so I'm glad you picked up on this. It does seem like a bit of an overload. But, there were two options:

* create setters/getters (`set_humidity`, `get_humidity`) and the same with temperature
* Have a single action that is capable of introspection as well as setting.

To date, I've been taking the first option, as it seems to read cleaner in a workflow if I were to describe it in human language.

I don't know what is going to make best sense really in terms of both usability now and in the future. Specifically how I might take logic decisions in a WF based on results. Certainly open to thoughts here.

packs/nest/actions/fan.yaml
jfryman: I also played with this pattern a few ways last night. A lot of them were to reduce the amount of code duplication (as a direct limitation to my knowledge of Python).

First option: function defaults. Minimal code, but hard-coded and does not give the user an easy way to globally override the default Structure/Device IDs.
Second option: Metaparameter defaults. No code, but lots of duplication in each file. Could drift
Third option: Global defaults. One stop shop, but logic overhead in each action to check if a value has been passed. But, tightly binds script to StackStorm via `config.txt`. I could abstract this a bit, but feels like it breaks the spirit of the action.

I sort of landed on the second option in both terms of speed and ease to write. Having a python action managing whether or not a default parameter or not seemed ugly, and something I wanted to bring up. I feel like this data binding should happen for free somewhere and have a data-binding convention on a per-action basis. This keeps the actions themselves focused on the lifting, and not logic to interpret and bootstrap the action.

packs/nest/config.yaml
```diff
@@ -0,0 +1,3 @@
+--- 
+  username: "steve@fryman.io"
+  password: "}gjYzi8Z84kf;(LirrmGfLHkR6b9tohaDUceETz4#KXc4Nunac"
```
manasdk: !

packs/nest/actions/fan.yaml
manasdk: Hahaha. Yeah sounds like a Nest.

packs/nest/config.yaml
jfryman: fuuuuuuu. I keep doing this. :/

manasdk:
Nest + hue. Next SmartThings?

With that I should have everything to automate things at home :)

packs/nest/actions/fan.yaml
manasdk: Btw, auto or "auto"

jfryman: Stylistically, not sure how best to approach. TIL that `on` in YAML is actually a `true` value, and casts as such when parsed. So, `"on"` for sure to keep the literal string, but sort of not necessary with "auto".

Certainly open to suggestions here.

jfryman:
@manasdk yup. I'll hopefully finish the SmartThings integration this week. Actions first, then I'll take another pass at sensors across all the actions.

I also have plans for Logitech Smart Control, so I can get IR blasting in here too. :)

packs/nest/actions/fan.yaml
Kami: I usually combine both options together - function defaults and parameter defaults in the metadata file.

It might be slightly confusing to the pack author / contributor if the default values get out of sync, but for the end user it doesn't really make a difference since parameter defaults override function defaults.

Function defaults are there to make code more clear, reusable and useful if the action is used outside the StackStorm action context.

packs/nest/actions/humidity.py
Kami: I personally prefer to be explicit and have 2 methods aka the approach one.

jfryman: @Kami but whyyyyyy? I cannot think of a compelling reason to go one way or another right now, other than simpler actions. That may be reason enough, but idk.

Kami: Since it's explicit - if you are calling setter it's clear you are setting a value, likewise for the getter. You don't need to go check / dig into the code to see what's going on.

This approach is also not mutually exclusive with returning a value which was passed to the setter inside the setter function.

packs/nest/actions/fan.yaml
jfryman: > Function defaults are there to make code more clear, reusable and useful if the action is used outside the StackStorm action context.

Great point. I can get behind that.

packs/nest/actions/humidity.py
jfryman: WFM. I think the truism we want to call out here is:

Be Explicit

This seems to be a common theme in our conversations internally and externally.

packs/nest/actions/fan.yaml
lakshmi-kannan: I prefer explicit quoting. I have seen some weird parse errors if the value is unquoted.

jfryman:
Alright... made the actions more explicit, and cleaned up how the default device is referenced. This is GTG.

PR #164, opened 2015-03-24
[Itxaka] Rabbitmq new sensor

Kami:
This pull request contains some fixes and improvements for rabbitmq sensor from #162.

Improvements / fixes:

1. Only call close() if connection is defined
2. Support for servers which don't require authentication


Kami:
There is another issue / limitation I have encountered.

Raw message body is rarely useful because most frameworks and libraries serialize the payload in one way or another (usually it's either json or pickle in combination with a some kind of compression).

To make this sensor even more useful, there are a couple of things we could do:

1. Try to automatically deserialize the message payload using two most common serialization formats - pickle and json. If it fails, simply return raw message payload (and yes, pickle is not really safe, but tell this to people who use pickle and not json as a serialization format :)).

2. Require user to explicitly specify serialization method used in the config and use this method to try to deserialize the body. This approach is less "magical" and more explicit, but it will only work if user only uses one kind of a serialization format.

enykeev:
> and yes, pickle is not really safe, but tell this to people who use pickle and not json as a serialization format

yeah, this guys for example https://github.com/StackStorm/st2/blob/8b84d304fc920bc0d432f88c4628c6f2101a191a/st2reactor/st2reactor/rules/worker.py#L30-L32

Jokes aside, I also prefer the second approach, but instead of defining it in config, this is another place where configured triggers may come handy. Define the format as trigger parameter and ignore the message that can't be serialized. Then, if user for some reason uses more than one format, he could create multiple rules to handle each of them individually.

Although, I have hard time imagining someone using different types of serialization withing the same exchange\queue. If configured triggers is no-go, how about defining serialization on per exchange\queue basis?

packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -0,0 +1,53 @@
...
+        self.channel.basic_ack(delivery_tag=method.delivery_tag)
+
+    def update_trigger(self, trigger):
+        pass
+
+    def add_trigger(self, trigger):
+        pass
+
+    def remove_trigger(self, trigger):
+        pass
```
enykeev: Do we have any rules about a single blank line at the end of each file?

Itxaka:
I think 2 would be the preferred way. Explicit all the way :D

RE: One message as a time. 
It was done on purpouse, as a sensor for small queues. For bigger queues this sensor is not really useful as it will only process one message at a time. This is where my comment below  makes sense.

Regarding the sensors, is there any way of launching several of the same sensor with different variables attached to them? Not useful in this case(Yet, but it could be for high/low traffic queues or process one/exhaust the queue modifiers) but I was thinking of doing a database sensor (Needed for a prototype im working on) that would get a query as parameter and launch a sensor with it, but it seems that rigth now its not possible. A sensor its a monolithic piece that only does one thing.

Is there any plans to expand sensors to be able to use parameters and have more than one of the same sensor running? Make them more dynamic?

Kami:
@Itxaka Alright, I will go with the approach 2 then :)

As far as your second questions go - we are just working on figuring this out. Some patterns have emerged for solving this issue (see https://gist.github.com/Kami/098c6f263115a1862899 for examples), but I'm not too happy with any of them. Because of that we are just brainstorming and trying to come up with a best, consistent and easy to use (for sensor authors) approach for solving this.

In short - we are leaning towards the solution which allows users to run multiple instances of a sensor - each one with a different config.

Itxaka:
@Kami Great to hear that there is work being done in that part. I think it would turn the sensors from useful to one of the best features of stackstorm. Adding sensors with different configs will make everything much easier to set up byhaving some base sensors!

Hopefully it comes soon!

packs/rabbitmq/sensors/queues_sensor.py
Kami: Yeah, pep8 calls for a single blank line between methods and 2 between top level class definitions...

Kami:
@enykeev Yeah, I agree, it would be weird to use different serialization formats for messages published to the same queue.

For now, I added a change where user can define this format globally for all the messages in the config. I will go ahead and merge changes into master.

Itxaka:
@Kami readme.md from the pack needs updating for the new config parameter and the main st2contrib readme.md needs updating too as rabbitmq only shows the actions.

Kami:
@Itxaka heh, I've actually updated the readme locally but forgot to add the file.

Anyway, pushed the change now :)

packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -0,0 +1,85 @@
...
+
+from st2reactor.sensor.base import PollingSensor
+
+DESERIALIZATION_FUNCTIONS = {
+    'json': json.loads,
+    'pickle': pickle.loads
+}
+
+
+class RabbitMQSensor(PollingSensor):
```
lakshmi-kannan: Commenting on this PR as well. We probably need a more specific class name. I am sure we'll write more rabbit integrations. (I am looking at you @manasdk :P)

PR #165, opened 2015-03-24
Add urbandict pack

Kami:
Just something so we can add some puns to our chat bot.

```bash
python ./st2client/st2client/shell.py run urbandict.get_definitions term=spottease
.....

id: 5511d3800640fd38d5e0f228
status: succeeded
result: 
{
    "result": "[{"word": "Spot Tease\n", "example": "\nAlright! Parking spot just two away from the mall! Wait...damn Smart Cars parked there! Stupid Spot Tease!!!\n", "def": "\nA parking spot that appears to be open, but is actually taken by a small car or motorcycle. Also can refer to a car that takes up 90% of its own spot and 10% of the one next to it, leaving no room for another car to park.\n"}]", 
    "exit_code": 0, 
    "stderr": "", 
    "stdout": ""
}

```

### DIFF
--- a/packs/urbandict/README.md
+++ b/packs/urbandict/README.md
@@ -1,0 +1,7 @@
+# Urban Dictionary Integration Pack
+
+Pack which allows integration with [Urban Dictionary](http://www.urbandictionary.com/).
+
+## Actions
+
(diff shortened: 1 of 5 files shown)


manasdk:
Why not! 

+1

PR #167, opened 2015-03-24
adds a new sensor for mmonit events and Actions for mmonit

Itxaka:
Will add some actions later

### DIFF
--- a/packs/mmonit/actions/get_uptime_host.yaml
+++ b/packs/mmonit/actions/get_uptime_host.yaml
@@ -13,6 +13,14 @@ parameters:
     type: integer
     description: 'The range id. Possible values: 0=today, 5=week, 6=month, 7=three months, 8=six months, 9=year and 12=custom range. Default is 0=today'
     required: false
+    enum:
+      - 0
+      - 5
+      - 6
+      - 7
+      - 8
+      - 9
+      - 12
   datefrom:
     type: integer
     description: 'A unix-timestamp (seconds since the epoch, January 1, 1970, 00:00:00 GMT). The timestamp specify the start of a custom uptime range. The range id must also be set to 12, otherwise this argument is ignored.'
--- a/packs/mmonit/actions/get_uptime_host.yaml
+++ b/packs/mmonit/actions/get_uptime_host.yaml
@@ -1,0 +1,23 @@
+---
+name: get_uptime_hosts
+runner_type: run-python
+description: Returns services uptime for a particular host.
+enabled: true
+entry_point: get_uptime_hosts.py
+parameters:
+  host_id:
+    type: integer
+    description: ID of the host
+    required: true
+  uptime_range:
+    type: integer
+    description: The range id. Possible values: 0=today, 5=week, 6=month, 7=three months, 8=six months, 9=year and 12=custom range. Default is 0=today
+    required: false
+  datefrom:
+    type: integer
+    description: A unix-timestamp (seconds since the epoch, January 1, 1970, 00:00:00 GMT). The timestamp specify the start of a custom uptime range. The range id must also be set to 12, otherwise this argument is ignored.
+    required: false
+  dateto:
+    type: integer
+    description: A unix-timestamp, specifying the end of a custom uptime range.
+    required: false
--- a/packs/mmonit/actions/get_uptime_host.yaml
+++ b/packs/mmonit/actions/get_uptime_host.yaml
@@ -7,17 +7,17 @@ entry_point: get_uptime_hosts.py
 parameters:
   host_id:
     type: integer
-    description: ID of the host
+    description: 'ID of the host'
     required: true
   uptime_range:
     type: integer
-    description: The range id. Possible values: 0=today, 5=week, 6=month, 7=three months, 8=six months, 9=year and 12=custom range. Default is 0=today
+    description: 'The range id. Possible values: 0=today, 5=week, 6=month, 7=three months, 8=six months, 9=year and 12=custom range. Default is 0=today'
     required: false
   datefrom:
     type: integer
-    description: A unix-timestamp (seconds since the epoch, January 1, 1970, 00:00:00 GMT). The timestamp specify the start of a custom uptime range. The range id must also be set to 12, otherwise this argument is ignored.
+    description: 'A unix-timestamp (seconds since the epoch, January 1, 1970, 00:00:00 GMT). The timestamp specify the start of a custom uptime range. The range id must also be set to 12, otherwise this argument is ignored.'
     required: false
   dateto:
     type: integer
-    description: A unix-timestamp, specifying the end of a custom uptime range.
+    description: 'A unix-timestamp, specifying the end of a custom uptime range.'
     required: false
--- a/packs/mmonit/actions/list_events.py
+++ b/packs/mmonit/actions/list_events.py
@@ -2,42 +2,43 @@ from lib.mmonit import MmonitBaseAction
 
 
 class MmonitListEvents(MmonitBaseAction):
-    def run(self, **kwargs):
+    def run(self, active=None, hostid=None, hostgroupid=None, servicenameid=None, servicegroupid=None, servicetype=None,
+            eventtype=None, state=None, datefrom=None, results=None, startindex=None, sort_key=None, sort_dir=None):
         self.login()
         data = {}
-        # Way too explicit for my taste but I guess its easier to understand
-        if 'active' in kwargs:
-            data['active'] = kwargs['active']
-        if 'hostid' in kwargs:
-            data['description'] = kwargs['hostid']
-        if 'hostgroupid' in kwargs:
-            data['hostgroupid'] = kwargs['hostgroupid']
-        if 'servicenameid' in kwargs:
-            data['servicenameid'] = kwargs['servicenameid']
-        if 'servicegroupid' in kwargs:
-            data['servicegroupid'] = kwargs['servicegroupid']
-        if 'servicetype' in kwargs:
-            data['servicetype'] = kwargs['servicetype']
-        if 'eventtype' in kwargs:
-            data['eventtype'] = kwargs['eventtype']
-        if 'state' in kwargs:
-            data['state'] = kwargs['state']
-        if 'datefrom' in kwargs:
-            data['datefrom'] = kwargs['datefrom']
-        if 'results' in kwargs:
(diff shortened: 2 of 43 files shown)


packs/mmonit/sensors/events_sensor.py
```diff
@@ -0,0 +1,57 @@
...
+            self._login()
+
+        events_list = self.session.get("{}/reports/events/list").json()
+        events = self._clear_list(events_list)
+        for event in events:
+            payload = {"host": event['hostname'], "id": event['id'],
+                       "event": event['event'], "servicename": event["servicename"],
+                       "full_data": event}
+            self._sensor_service.dispatch(trigger="mmonit.new_event", payload=payload)
+            self._sensor_service.set_value('monit.last_event', event["id"])
```
manasdk: Better if the last_event is saved at the end for the last_event that is processed.

packs/mmonit/sensors/events_sensor.yaml
```diff
@@ -0,0 +1,22 @@
...
+      name: "new_event"
+      description: "Trigger which indicates that a new event has arrived at mmonit"
+      payload_schema:
+        type: "object"
+        properties:
+          host:
+            type: "string"
+          id:
+            type: "integer"
+          event:
```
manasdk: Is the set of possible values of event known upfront? If so instead of a wrapper new_event you could also choose to break it out. Would make it easier to consume in a rule.

packs/mmonit/sensors/events_sensor.py
Itxaka: That has the side effect of triggering duplicated events if something goes wrong. So you have 50 events, you are going trougth the number 28, somebody restarts the sensor service or it dies for some reason, then your last_event stored was not the 28 but something complete different and the sensor will trigger those events again.

packs/mmonit/sensors/events_sensor.yaml
Itxaka: Can you elaborate on this?
The possible values are know, yes. Do you mean passing the whole object? Does that allow for rule filtering based on fields?

packs/mmonit/sensors/events_sensor.py
manasdk: Sorta hoping we could ignore that problem but yes we would have a hole in that case. This is accurate but ends up making quite a few calls into the K-V store. As of now I don't know if that will be an issue we shall find out :)

Ok with your approach.

packs/mmonit/sensors/events_sensor.yaml
manasdk: I presume that event['event'] has a defined set of values. Each of those could be defined as a separate TriggerType.

say event['event'] values are one of ['event_x', 'event_y', 'event_z']. Then you can map these to 3 triggertypes. So instead of 'new_event' the TriggerTypes would be 'event_x', 'event_y' and 'event_z'.

In a rule instead of referring to 'new_event' in the trigger field you can now refer to the 3 other types.

Again this is all dependent on if the values are know, the values are a small enough number and there is a requirement to perform different work on each of these events. It will provide some more visibility to a rule author.

Itxaka: Well, there is quite a few values possible so that why I set up only the 3 that looked more importatn for filtering and left the rest in the whole object. 

Ill have a look tonight to see how many there are.

manasdk: It cool. If its more than a handful then what you have works. There is always some design tension in this area about 1 event that multiplexes multiple or several events - I feel that the multiplexing option works in this case.

packs/mmonit/sensors/events_sensor.py
```diff
@@ -0,0 +1,57 @@
...
+        if login.status_code != 200:
+            raise Exception("Could not login to mmonit {}".format(login.reason))
+
+    def poll(self):
+        # always check for status_code on a secured page on each poll in case we get logged out
+        test_login = self.session.get("{}/session/get".format(self.url))
+        if test_login.status_code != 200:
+            self._login()
+
+        events_list = self.session.get("{}/reports/events/list").json()
```
lakshmi-kannan: I think this is missing a  ".format(self.url)"

packs/mmonit/sensors/events_sensor.py
```diff
@@ -0,0 +1,57 @@
...
+        return events['records']
+
+    def add_trigger(self, trigger):
+        pass
+
+    def update_trigger(self, trigger):
+        pass
+
+    def remove_trigger(self, trigger):
+        pass
```
lakshmi-kannan: Nit: newline.

packs/mmonit/sensors/events_sensor.yaml
Itxaka: Oh I get it now. The problem is the events could be any type actually, I was thinking about fields instead. Unfortunately the event type depends on the type of check monit does (process, url, mysql, ssh, file check, checksum,....) and there are way too many types to set them in there, at least in a consistent manner.

packs/mmonit/sensors/events_sensor.py
Itxaka: Good catch!

Itxaka:
@manasdk @lakshmi-kannan @Kami I guess is ready for a review and merge if everything is ok. Most of the API of mmonit is included, only missing the groups management and user management part, which I may do later, still dont know if they have a part in an automated system (groups kind of have, user management I dont know)

packs/mmonit/actions/list_events.py
```diff
@@ -0,0 +1,43 @@
+from lib.mmonit import MmonitBaseAction
+
+
+class MmonitListEvents(MmonitBaseAction):
+    def run(self, **kwargs):
+        self.login()
+        data = {}
+        # Way too explicit for my taste but I guess its easier to understand
```
Kami: Instead of using kwargs, can you please explicitly declare all the arguments in the method signature?

kwargs breaks method instrospection and many other things.

packs/mmonit/actions/list_events.yaml
```diff
@@ -0,0 +1,63 @@
...
+    type: integer
+    description: 'Show only events for the given service name id.'
+    required: false
+  servicegroupid:
+    type: integer
+    description: 'Show only events for the given servicegroup id.'
+    required: false
+  servicetype:
+    type: integer
+    description: 'Show only events for the given service type. Possible values: 0=filesystem, 1=directory, 2=file, 3=process, 4=remote host, 5=system, 6=fifo, 7=program.'
```
Kami: If there is a small set of possible values which doesn't changes, there is an "enum" attribute you can use.

See the following link for an example - https://github.com/StackStorm/st2contrib/blob/master/packs/jira/actions/create_issue.yaml#L12

packs/mmonit/actions/list_hosts.py
```diff
@@ -0,0 +1,14 @@
+from lib.mmonit import MmonitBaseAction
+
+
+class MmonitListHosts(MmonitBaseAction):
+    def run(self, host_id):
+        self.login()
```
Kami: If you wanted to save a couple of lines of repeated code across the actions, you could create a custom context manager class which logs in on `__enter__` and logs out on `__exit__` and then do something like this:

```python
with MoninSession() as session:
    session.get(...)
```

packs/mmonit/actions/list_status_hosts.py
```diff
@@ -0,0 +1,30 @@
+from lib.mmonit import MmonitBaseAction
+
+
+class MmonitListStatusHost(MmonitBaseAction):
+    def run(self, host_id=None, hostgroupid=None, status=None, platform=None, machine=None, led=None):
+        self.login()
+        data = {}
+
+        # Way too explicit for my taste but I guess its easier to understand
```
Kami: To clarify it, that's not what I meant with explicit :)

With explicit, I simply meant explicitly declaring all the arguments in the method signature (which is indeed done here).

packs/mmonit/actions/update_host.py
```diff
@@ -0,0 +1,37 @@
...
+            data['monitpassword'] = monitpassword
+        if skew is not None:
+            data['skew'] = skew
+
+        req = self.session.post("{}/admin/hosts/update".format(self.url), data=data)
+
+        try:
+            return req.json()
+        except Exception as error:
+            return error.message
```
Kami: On a fatal exception we should re-throw it (or if you don't do anything with the exception as in this example, just remove the except clause) otherwise the action will be considered as succeeded and not as failed.

In short, action is considered as failed if one of those conditions occurs:

* Exception is thrown
* Script exists with a non zero code

In the `python-runner` case, you should always aim to throw an exception with a meaningful error message when possible.

packs/mmonit/requirements.txt
```diff
@@ -0,0 +1 @@
+requests
```
Kami: To be on the safe side we should pin in to a specific set of versions (e.g. `>=2.6.0,<3.0` should work).

Kami:
@Itxaka Thanks, added somre more comments.

Itxaka:
Will update later, thanks @kami for your comments :)

packs/mmonit/actions/list_hosts.py
Itxaka: Great idea!

Kami: To clarify it - this is a nice to have, but not a blocker :)

Itxaka:
@manasdk @lakshmi-kannan @Kami Updated with fixed for last comments :)

Kami:
Thanks.

Changes look good so I went ahead and merged the PR :)

packs/mmonit/actions/get_uptime_host.yaml
```diff
@@ -0,0 +1,31 @@
...
+parameters:
+  host_id:
+    type: integer
+    description: 'ID of the host'
+    required: true
+  uptime_range:
+    type: integer
+    description: 'The range id. Possible values: 0=today, 5=week, 6=month, 7=three months, 8=six months, 9=year and 12=custom range. Default is 0=today'
+    required: false
+    enum:
```
Kami: It might be a bit more user-friendly to use actual string values here and in other places where enums are used (today, week, month, etc.) instead of the numeric values. Then we could map those string values to numeric values in the code.

Anyway, that's not a blocker or anything, just an optimization we can do in the future :)

PR #168, opened 2015-03-26
Initial commit of Trello pack

jfryman:
This PR introduces StackStorm support for Trello, an online task manager.

## Supported Actions
```
+---------------------------+--------+--------------------+------------------------------------------------------+
| ref                       | pack   | name               | description                                          |
+---------------------------+--------+--------------------+------------------------------------------------------+
| trello.add_board          | trello | add_board          | Create a new board                                   |
| trello.add_card           | trello | add_card           | Add a new card to a list                             |
| trello.add_list           | trello | add_list           | Add a new list to a board                            |
| trello.close_board        | trello | close_board        | Close a board                                        |
| trello.close_card         | trello | close_card         | Close a card                                         |
| trello.close_list         | trello | close_list         | Close a list belonging to a board                    |
| trello.find_board_by_name | trello | find_board_by_name | Lookup a board ID based on name. Returns one or more |
|                           |        |                    | IDs                                                  |
| trello.find_card_by_name  | trello | find_card_by_name  | Lookup a Card ID based on name. Returns one or more  |
|                           |        |                    | IDs                                                  |
| trello.find_list_by_name  | trello | find_list_by_name  | Lookup a list ID based on name. Returns one or more  |
|                           |        |                    | IDs                                                  |
| trello.move_card          | trello | move_card          | Move a card from one board/list to another           |
|                           |        |                    | board/list                                           |
| trello.view_boards        | trello | view_boards        | Return a dictionary of all boards and their IDs      |
| trello.view_cards         | trello | view_cards         | View all cards on a board                            |
| trello.view_lists         | trello | view_lists         | View all lists belonging to a board                  |
| trello.view_organizations | trello | view_organizations | List all organizations for user                      |
+---------------------------+--------+--------------------+------------------------------------------------------+
```

### DIFF
--- a/packs/trello/README.md
+++ b/packs/trello/README.md
@@ -1,0 +1,48 @@
+# Trello Integration Pack
+
+Integration pack that provides support for Trello, an online Task Tracking tool
+
+## Configuration
+
+* `api_key` - User API Key
+* `token` - User oAuth Token for R/W Access
+
+Note: Each action in this pack also takes `api_key` and `token` as parameters.
+If provided, actions will prefer the runtime credentials over the system provided
+credentials.
+
+### Obtaining Credentials
+#### API Token
+
+API Token can be retrieved from https://trello.com/app-key while logged into your
+account.
+
+#### oAuth Token
+
+To obtain an oAuth token, refer to the documentation at https://trello.com/docs/gettingstarted/#getting-a-token-from-a-user
+
+## Supported Actions
+```
++---------------------------+--------+--------------------+------------------------------------------------------+
+| ref                       | pack   | name               | description                                          |
++---------------------------+--------+--------------------+------------------------------------------------------+
+| trello.add_board          | trello | add_board          | Create a new board                                   |
+| trello.add_card           | trello | add_card           | Add a new card to a list                             |
+| trello.add_list           | trello | add_list           | Add a new list to a board                            |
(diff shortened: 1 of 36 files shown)


jfryman:
Some commentary:

Note the `api_key` and `token` attributes that are part of basically every action. The reasoning here comes later with ChatOps. We don't want a single user taking credit for each action that is done via ChatOps, so by having the ability to override on a per-action basis we can keep user integrity. (Stored via K/V store or otherwise). This pattern though... feels ugly. I have to do some weird overwrite logic in `lib/action.py`, and then have to call the `_set_credentials()` attribute. I wonder if there is a pattern here to make this cleaner.

I tried overwriting the constructor as a pattern, but that wouldn't work due to how attributes are passed into the python runner.

lakshmi-kannan:
Yeah, this is an interesting problem. It would be cool to make these parameters attributes of the class and if the base class actually set these attributes run time. I don't see a better way of doing this with current infrastructure other than what you did.

Kami:
Yeah, that's a similar pattern I've also followed in some other packs - specify global defaults in the config and allow user to override it on per action basis.

Imo, that's a reasonable pattern. As far as code goes, another approach would be to do some "magic" with a decorator, but I prefer an explicit approach over a decorator.

PR #169, opened 2015-03-29
[st2] Add new kv.set_object and kv.get_object action

Kami:
This pull request adds new actions which allow users to store and retrieve objects from a datastore.

Internally, those objects are serialized as JSON.

```bash
st2 run st2.kv.set_object key=keyo value='{"foo": "bar"}'
...
id: 551825cd0640fd35173960e7
status: succeeded
result: 
{
    "result": {
        "key": "keyo", 
        "value": {
            "foo": "bar"
        }
    }, 
    "exit_code": 0, 
    "stderr": "", 
    "stdout": ""
}

st2 run st2.kv.get key=keyo
...
id: 551826170640fd35173960eb
status: succeeded
result: 
{
    "result": "{"foo": "bar"}", 
    "exit_code": 0, 
    "stderr": "", 
    "stdout": ""
}

st2 run st2.kv.get_object key=keyo
...
id: 551826010640fd35173960e9
status: succeeded
result: 
{
    "result": {
        "foo": "bar"
    }, 
    "exit_code": 0, 
    "stderr": "", 
    "stdout": ""
}
```

### DIFF
--- a/packs/st2/actions/kv_set_object.yaml
+++ b/packs/st2/actions/kv_set_object.yaml
@@ -1,0 +1,13 @@
+---
+name: 'kv.set_object'
+enabled: true
+description: 'Serialize and store object in a datastore'
+runner_type: run-python
+entry_point: kv_set_object.py
+parameters:
+  key:
+    required: True
+    type: string
+  value:
+    required: True
+    type: object
--- a/packs/st2/README.md
+++ b/packs/st2/README.md
@@ -1,13 +1,24 @@
 # StackStorm Integration Pack
 
-The super-meta package! This integration allows integration with StackStorm. 
+The super-meta package! This integration allows integration with StackStorm.
 
-Requires StackStorm <= `v0.8.0`
+Requires StackStorm >= `v0.8.0`
 
 ## Configuration:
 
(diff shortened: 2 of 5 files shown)


packs/st2/actions/kv_set_object.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: 'Serialize and store object in a datastore'
+runner_type: run-python
+entry_point: kv_set_object.py
+parameters:
+  key:
+    required: True
+    type: string
+  value:
+    required: True
+    type: object
```
lakshmi-kannan: Does this work with primitives? Someone from IRC complained it doesn't.

lakshmi-kannan:
Nice!

packs/st2/actions/kv_set_object.yaml
lakshmi-kannan: Oh. NVM. set_object.

Kami: Yep, it's only object, but those objects (dicts) can contain arbitrary types which are supported in JSON :)

PR #170, opened 2015-03-29
[twitter] New update_status action.

Kami:


Kami:
Just FYI - keys on the screenshot are revoked.

packs/twitter/actions/update_status.yaml
```diff
@@ -0,0 +1,11 @@
+---
+  name: "update_status"
+  runner_type: "run-python"
+  description: "Update your status (post a new tweet)."
+  enabled: true
+  entry_point: "update_status.py"
+  parameters:
+    status:
+      type: "string"
```
lakshmi-kannan: UTF-8 supported?

lakshmi-kannan:
LGTM

packs/twitter/actions/update_status.yaml
Kami: Yep, Twitter supports unicode, but I believe jsonschema only has "string" type and Python json schema parser we use handles unicode just fine, right?

lakshmi-kannan: That is my question, dear sir! Can you please validate?

Kami: Yep, looks like it does:

* https://github.com/Julian/jsonschema/blob/master/jsonschema/validators.py#L52
* https://github.com/Julian/jsonschema/blob/master/jsonschema/compat.py#L30

PR #174, opened 2015-04-02
Do not expect name profile info in Slack sensor

ryandub:
`first_name`, `last_name`, and `real_name` do not always exist for every
Slack user (bots maybe?). Expecting these values causes the sensor to
crash with a `KeyError`. This replaces them with 'Unknown' if they are not
found which allows the sensor to continue running.


Kami:
Heh, thanks. I just pushed some improvements to the sensor 30 mins or so ago - #173 

I will look into your changes shortly.

ryandub:
Oh awesome! I've seen the duplicated messages a few times too. Great to see a fix for that. Thanks! :)

Kami:
Looks good. Merged. Thanks.

PR #176, opened 2015-04-06
Add private group support for slack sensor.

Gifflen:
Previously messages in private groups were ignored due to slack treating them slightly differently than normal channels. This adds support to process private group messages and cache available groups as we do channels.


Kami:
Thanks.

I will look into it and test it and get back to you soon.

Kami:
Changes look good and also everything worked fine during testing.

Merged into trunk. Thanks.

PR #177, opened 2015-04-07
[azure] New azure pack

Kami:
It comes with an initial set of compute and storage actions.

```bash
+----------------------+-------+----------------------+----------------------+
| ref                  | pack  | name                 | description          |
+----------------------+-------+----------------------+----------------------+
| azure.create_contain | azure | create_container     | Create a new storage |
| er                   |       |                      | container.           |
| azure.create_vm      | azure | create_vm            | Create a new VM.     |
| azure.delete_contain | azure | delete_container     | Delete a storage     |
| er                   |       |                      | container.           |
| azure.delete_object  | azure | delete_object        | Delete an object.    |
| azure.destroy_vm     | azure | destroy_vm           | Destroy a VM.        |
| azure.list_container | azure | list_container_objec | List storage objects |
| _objects             |       | ts                   | for the provided     |
|                      |       |                      | container.           |
| azure.list_container | azure | list_containers      | List storage         |
| s                    |       |                      | containers.          |
| azure.list_vms       | azure | list_vms             | List available VMs.  |
| azure.reboot_vm      | azure | reboot_vm            | Reboot a running VM. |
| azure.upload_file    | azure | upload_file          | Upload a file to the |
|                      |       |                      | provided container.  |
+----------------------+-------+----------------------+----------------------+

```

### DIFF
--- a/packs/azure/actions/lib/formatters.py
+++ b/packs/azure/actions/lib/formatters.py
@@ -1,0 +1,16 @@
+__all__ = [
+    'to_node_dict'
+]
+
+
+def to_node_dict(node):
+    result = {
+        'id': node.id,
+        'name': node.name,
+        'state': node.state,
+        'public_ips': node.public_ips,
+        'private_ips': node.private_ips,
+        'extra': node.extra or {}
+    }
+
+    return result
--- a/packs/azure/actions/lib/formatters.py
+++ b/packs/azure/actions/lib/formatters.py
@@ -1,5 +1,7 @@
 __all__ = [
-    'to_node_dict'
+    'to_node_dict',
+    'to_container_dict',
+    'to_object_dict'
 ]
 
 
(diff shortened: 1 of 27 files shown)


packs/azure/actions/lib/formatters.py
```diff
@@ -0,0 +1,16 @@
+__all__ = [
+    'to_node_dict'
+]
+
+
+def to_node_dict(node):
```
Kami: We need to come up with a good pattern / solution for handling common code which is needed in multiple packs which doesn't result in us copy and pasting code all over the place.

Currently, the same code is needed in libcloud, softlayer and azure pack.

I was thinking about creating a new Python package - `st2contrib-libcloud-formatters` or something along those lines, but I'm not too happy with this approach.

Feedback and suggestions are welcome.

Kami:
Also added some storage actions and updated the description.

packs/azure/actions/lib/formatters.py
manasdk: Create an egg and install taht egg in all virtualenvs?

lakshmi-kannan: We can create /opt/stackstorm/libs/ and throw these common code in there. When we create virtualenvs for actions or sensors, we can create symlinks or just add this path to PYTHONPATH.

Kami: That's a possibility, but it doesn't (and probably shouldn't) be available to all the packs. That's why I was thinking about going with a package which is explicitly added to `requirements.txt` inside the packs which need it.

lakshmi-kannan: Can pip install dependencies from a file location? If so, I like the requirements approach.

PR #178, opened 2015-04-07
Update Dripstat class name to less confusing name

jfryman:
This commit changes the class name to `dripstat`, which allows a user to reference this trigger by a much shorter, consumable name.


Kami:
@jfryman This will break the sensor without the matching code change (sensor class rename). We use this attribute to look for the class in the sensor file which is also specified in the metadata file.

Kami:
Also, sensor class name has nothing to do with the trigger name - trigger name is defined further down in the metadata file (`alert` in this case). and the full trigger reference is `<pack name>`.<trigger name>` which in this case is `dripstat.alert`.

PR #179, opened 2015-04-07
Do not crash Slack sensor if unexpected message

### LINKED ISSUE #175: Slack Sensor Instability
The messages coming through Slack's RTM API seem to be quite varied and may or may not contain certain data. For example, #174 addresses some of the variations in profile data. However, I have also seen some quirkiness with `data` not containing a `channel` key as well (I haven't dug in to find what these messages are yet, but that's weird - maybe private groups?). Here's a stacktrace from such an occurance:

```
  File "/usr/lib/python2.7/dist-packages/st2reactor/container/sensor_wrapper.py", line 336, in run
    self._sensor_instance.run()
  File "/usr/lib/python2.7/dist-packages/st2reactor/sensor/base.py", line 92, in run
    self.poll()
  File "/data/st2contrib/packs/slack/sensors/slack_sensor.py", line 45, in poll
    self._handle_result(result=result)
  File "/data/st2contrib/packs/slack/sensors/slack_sensor.py", line 75, in _handle_result
    handler_func(data=item)
  File "/data/st2contrib/packs/slack/sensors/slack_sensor.py", line 87, in _handle_message
    channel_info = self._get_channel_info(channel_id=data['channel'])
  File "/data/st2contrib/packs/slack/sensors/slack_sensor.py", line 122, in _get_channel_info
    channel=channel_id)['channel']
KeyError: 'channel'
```
Note: Line numbers may be off as I was poking around on this file a bit...

Ideally, I would like for the sensor to keep on trucking despite running into issues with a particular message. Would it be appropriate to wrap `self._handle_result(result=result)` in the [poll](https://github.com/StackStorm/st2contrib/blob/master/packs/slack/sensors/slack_sensor.py#L51) function in a `try`/`except` block? On error, it could log the event, but keep the sensor running.

Thoughts?

ryandub:
Closes #175

If the message from Slack is not formatted as expected, log the issue
and continue - do not crash sensor.

### DIFF
--- a/packs/slack/sensors/slack_sensor.py
+++ b/packs/slack/sensors/slack_sensor.py
@@ -30,7 +30,7 @@ class SlackSensor(PollingSensor):
         self._strip_formatting = self._config['sensor'].get('strip_formatting',
                                                             False)
         self._handlers = {
-            'message': self._handle_message,
+            'message': self._handle_message_ignore_errors,
         }
 
         self._user_info_cache = {}
@@ -183,6 +183,14 @@ class SlackSensor(PollingSensor):
         }
 
         self._sensor_service.dispatch(trigger=trigger, payload=payload)
+
+    def _handle_message_ignore_errors(self, data):
+        try:
+            self._handle_message(data)
+        except Exception as exc:
+            self._logger.info("Slack sensor encountered an error "
+                              "handling message: %s" % exc)
+            pass
 
     def _get_user_info(self, user_id):
         if user_id not in self._user_info_cache:


Kami:
Thanks.

For readability reasons, I would prefer to add another method which calls `self._handle_message` and catch the exception there (or something along those lines - decorator would also work).

For example:

```python
...
self._handlers = {
   'message': self._handle_message_ignore_errors,
}
....

def handle_message_ignore_errors(...):
  try:
    self._handle_message(...)
  except:
   ...
```

ryandub:
Thanks @Kami ! I've updated with your suggestion.

jfryman:
Awesome! Thanks so much for the code! :metal:

Kami:
@ryandub Thanks again.

This pull request and changes in #181 should now make sure the sensor doesn't die on non-fatal errors.

PR #187, opened 2015-04-24
chef pack updates

dennybaa:
Forgot to merge before, there's a Readme)


packs/chef/actions/lib/shellhelpers.py
```diff
@@ -8,8 +8,7 @@
 
 # Starts command in subprocess processing output as it appears.
 def shell_out(command, env=None):
-    env = env or {}
-    env = os.environ.copy().update(env)
```
dennybaa: @Kami btw why do you prefer
```python
env = env or {}
env = os.environ.copy().update(env)
```
to
```
env = os.environ.copy().update(env or {})
```

However the former got occasionally with my commit. In future going to use your notation. But tell, please, why it's preferable.

PR #189, opened 2015-04-28
RFR - Add auth headers to st2 API calls

lakshmi-kannan:
With 0.9 release of st2, API endpoints could have authentication turned on. Therefore, nagios and sensu handlers need to use auth headers. This PR adds the header. 

An auth token for st2 can be generated using the st2 CLI. 
http://docs.stackstorm.com/latest/authentication.html#testing

TODO:

* [X] - Test the handler.   
* [X] - https://github.com/StackStorm/st2/pull/1412

### DIFF
--- a/packs/nagios/etc/st2service_handler.py
+++ b/packs/nagios/etc/st2service_handler.py
@@ -12,7 +12,7 @@ import requests
 import yaml
 
 # ST2 configuration
-ST2_CONFIG_FILE = './config.yml'
+ST2_CONFIG_FILE = './config.yaml'
 
 ST2_API_BASE_URL = 'http://localhost:9101/v1'
 ST2_AUTH_BASE_URL = 'http://localhost:9100'
@@ -48,7 +48,13 @@ def _create_trigger_type():
             'description': 'Trigger type for nagios event handler.'
         }
         # sys.stdout.write('POST: %s: Body: %s\n' % (url, payload))
-        post_resp = requests.post(url, data=json.dumps(payload))
+        headers = {}
+        headers['Content-Type'] = 'application/json; charset=utf-8'
+
+        if ST2_AUTH_TOKEN:
+            headers['X-Auth-Token'] = ST2_AUTH_TOKEN
+
+        post_resp = requests.post(url, data=json.dumps(payload), headers=headers)
     except:
         sys.stderr.write('Unable to register trigger type with st2.')
         raise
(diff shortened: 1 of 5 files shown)


jfryman:
I really don't like this.

We're hell-bent on shipping auth, but we need this feature in order to not introduce regressions. I don't like we have less than 24 hours to battle test this before we release.

Seems like a blocker to me for shipping auth, imo.

jfryman:
Let me be clear...  the feature itself is fine. Just the timing and robustness is suspect to me

packs/nagios/etc/st2service_handler.py
```diff
@@ -130,6 +150,17 @@ def main(args):
 
 
 if __name__ == '__main__':
+    if not os.path.exists(ST2_CONFIG_FILE):
+        print('Configuration file not found. Exiting.')
```
Kami: Looks like this is a new and backward incompatible change, right?

Kami: If so, we should mention this in the docs in the sensu section.

packs/nagios/etc/st2service_handler.py
```diff
@@ -5,16 +5,24 @@
...
 import sys
-import urlparse
+
+import requests
+import yaml
 
 # ST2 configuration
-ST2_HOST = 'localhost'
-ST2_API_PORT = '9101'
+ST2_CONFIG_FILE = './config.yaml'
```
Kami: Typo, yml - yaml :P

lakshmi-kannan: Yeah dude. Spotted during testing and fixed in future commits. Hold your horses.

packs/nagios/etc/st2service_handler.py
```diff
@@ -5,16 +5,24 @@
     import simplejson as json
 except ImportError:
     import json
-import requests
+import os
 import sys
-import urlparse
+
+import requests
```
manasdk: is there a requirements.txt that goes with this pack that needs to be updated?

lakshmi-kannan: Oh yes. Forgot about pyyaml.

Kami: I was wondering about that as well - it looks like there isn't, but requests was already used before that change.

If nothing else, we need to mention it in the docs (if we don't do that already).

packs/nagios/etc/st2service_handler.py
```diff
@@ -74,14 +110,12 @@ def _register_with_st2():
 
 
 def _get_st2_triggers_url():
-    url = urlparse.urlunparse(('http', ST2_HOST + ':' + ST2_API_PORT, ST2_TRIGGERS_PATH,
-                              None, None, None))
+    url = '%s/%s' % (ST2_API_BASE_URL, ST2_TRIGGERS_PATH)
```
manasdk: use urljoin. This saves having to deal with the trailing backslash.

packs/nagios/etc/st2service_handler.py
lakshmi-kannan: Already added in the docs @Kami. You +1ed it. https://github.com/StackStorm/st2/pull/1412

PR #194, opened 2015-05-08
Add Use Case example to Twitter pack readme [skip ci]

armab:
I found @Kami article describing how to use Twitter pack with Slack http://stackstorm.com/2014/12/22/monitor-twitter-and-fire-automations-based-on-twitter-keywords-using-stackstorm/ extremely useful.

Have no idea if you have any strict rules about documentation (maybe those readmes are parsed and then reused somewhere else), but I think this HowTo link should be definitely here.

### DIFF
--- a/packs/twitter/README.md
+++ b/packs/twitter/README.md
@@ -2,6 +2,9 @@
 
 Pack which allows integration with Twitter.
 
+## Use cases
+* [HOWTO: Broadcast Twitter mentions about your Company to Slack channel](http://stackstorm.com/2014/12/22/monitor-twitter-and-fire-automations-based-on-twitter-keywords-using-stackstorm/)
+
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


armab:
I love such cases because they are about popularization.
With simple steps everyone can start without digging deep into the docs and just show first real benefits for business. I believe more entry points should look this way, showing simplicity.

---
This is one of those awesome cases when developer in 3 guys and a dog company can go to CEO and convince him: 
> Hey, you know _this is_ possible? Let's try ChatOps.

funny pic

Kami:
@armab 

Thanks. I think including links to the use case and articles which describes how you can use the pack is fine so I will go ahead and merge the PR :)

dzimine:
Let's turn that to a tutorial (may be in the docs = my pref, may be linked from the docs)

PR #195, opened 2015-05-08
[st2] Add actions for common operations on st2 resources, add ChatOps aliases

Kami:
This pull request adds support for ChatOps alisases which allow users to introspect available actions, sensors and more via chat.

TODO:

- [x] ChatOps command for execution re-run
- [ ] Fix "{{" escaping issue in st2
- [x] API endpoint for re-running an execution


packs/st2/actions/actions_list.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
Kami: Once we move st2client in a separate repo and refactor the code so it doesn't abuse kwargs so we can introspect the code, we can auto generate most of those actions.

manasdk:
move to cliff and we can reuse most of the openstack autogen .. most not all

packs/st2/actions/actions_list.yaml
lakshmi-kannan: LOL. I was about to talk about that.

packs/st2/aliases/actions_list.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name: "st2-actions-list"
```
manasdk: I don't think we should chatops this one. What is the point of showing stackstorm has these actions available they cannot be used.

manasdk:
I am very confused by the point of this pack - why the convoluted approach of going through aliases?

aliasexecution -> actionexecution -> cli -> actionexecution -> triggerinstance -> notification

huh?

packs/st2/aliases/actions_list.yaml
jfryman: Agreed. Aliases are the only things that are exposed for commands.

manasdk:
Will make sure that this change works well with what I am working on which is better output formatting.

packs/st2/actions/chatops_format_list_result.py
```diff
@@ -0,0 +1,34 @@
+from st2actions.runners.pythonrunner import Action
+
+from prettytable import PrettyTable
+
+__all__ = [
+    'St2ChatOpsFormatListResult'
+]
+
+
+class St2ChatOpsFormatListResult(Action):
```
jfryman: Nice. This is similar to what I was thinking with formatting. The only downside is that its another action to handle this, as opposed to something the client handles.

packs/st2/actions/executions_list.yaml
```diff
@@ -0,0 +1,15 @@
...
+enabled: true
+description: "Retrieve a list of executions."
+runner_type: run-python
+entry_point: executions_list.py
+parameters:
+  action:
+    type: "string"
+    description: "Optional action to filter the executions on"
+    required: false
+  status:
```
jfryman: `enum` opportunity?

Kami:
@manasdk I believe this is the same discussion as the one we had yesterday :)

It follows exactly the same flow as any other action which would be exposed through ChatOps would. And I believe we also already talked multiple times about why we need the workflow and formatter action - currently we don't have any native formatting support so there is no better or nicer way to do that.

And honestly, I'm quite happy with this approach. It follows and re-uses existing model and constructs. It doesn't introduce any one-offs or anything like that. When and if we add support for better result formatting, we can also simply port those actions to the new approach.

packs/st2/actions/executions_list.yaml
Kami: Yeah, I was thinking about it.

I think in this case it's fine to push a bit work to the API - this way it will be less maintenance and code which needs to be updated if we introduce a new status (which, I guess, it's unlikely at this point, but who knows :D)

PR #196, opened 2015-05-08
WIP: Docker build image action fixes

lakshmi-kannan:
- Print complete traceback on exceptions for better visibility.
- JSON output from docker daemon is inconsistent depending on whether
  a base image is already existing or is being downloaded. You can either
  get {'stream': 'some message'} or {'status': 'STATUS', other fields...}.
  So this action now prints the entire JSON output.
- While using chunked transfer (stream=True), sometimes we get a ReadTimeoutError.
  In most cases, it is temporary because docker daemon is loaded. So
  we just ignore the ReadTimeoutError and retry. Note this is not ideal
  but as a client there is little we could do.

### DIFF
--- a/packs/docker/actions/lib/docker_wrapper.py
+++ b/packs/docker/actions/lib/docker_wrapper.py
@@ -2,6 +2,10 @@ import sys
 import traceback
 
 from requests.packages.urllib3.exceptions import ReadTimeoutError
+try:
+    import simplejson as json
+except ImportError:
+    import json
 import six
 import docker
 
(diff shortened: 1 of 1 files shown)


packs/docker/actions/lib/docker_wrapper.py
```diff
@@ -1,10 +1,7 @@
 import sys
+import traceback
 
-try:
-    import simplejson as json
-except:
-    import json
-
+from requests.packages.urllib3.exceptions import ReadTimeoutError
```
lakshmi-kannan: I am not explicitly adding this to requirements but docker-py does depend on this. I can go either ways.

lakshmi-kannan:
I tested with this patch a bunch of times and I am able to build an image fine.

lakshmi-kannan:
Closing. Not working on this.

PR #198, opened 2015-05-11
Elasticsearch pack release 0.1.0

dennybaa:


Kami:
Thanks, I will look into the PR shortly.

packs/elastic/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Action | Description
+------ | -----------
+**alias** | Add indices to or remove them from aliases.
+**allocation** | Set routing allocation based on tags.
+**bloom** | Disable the bloom filter cache for indices.
+**close** | Close indices.
+**delete.indices** | Delete indices.
+**delete.snapshots** | Delete snapshots.
+**open** | Open indices.
```
Kami: I personally prefer to use more explicit names, especially since we could add other non-index related actions later on (e.g. delete document, get document, etc.).

Namespacing the action names into something like `elasticsearch.index.optimize`, `elasticsearch.documents/search.query`, etc. is probably better.

packs/elastic/actions/lib/static_metagen.py
```diff
@@ -0,0 +1,122 @@
...
+            if 'alias' in manifest:
+                alias_name = manifest.pop('alias')
+                alias_manifest = manifest.copy()
+                alias_manifest['name'] = alias_name
+                yield alias_manifest
+            yield manifest
+
+
+metagen = StaticMetagen()
+metagen.generate_from_file('lib/curator_actions.yaml')
```
Kami: Nice, I like that.

It would also be good to add a section for authors / developers to the readme which documents what this script does and how to use it to generate metadata files.

packs/elastic/actions/lib/static_metagen.py
```diff
@@ -0,0 +1,122 @@
...
+        fh = open('{0}_{1}.yaml'.format(module_type, action), 'w')
+        fh.write('---\n')
+        fh.write(yaml.dump(manifest, default_flow_style=False))
+        fh.close()
+
+
+    def generate_from_file(self, meta_file):
+        if meta_file is None:
+            return None
+        fh = open(meta_file)
```
Kami: You never close the original file.

Using something like `with` context manager is better and safer since you don't need to manually make sure to close the file.

packs/elastic/actions/lib/utils.py
```diff
@@ -0,0 +1,68 @@
...
+def check_master(client, master_only=False):
+    """
+    Check if master node.  If not, exit with error code
+    """
+    if master_only and not is_master_node(client):
+        logger.info('Master-only flag detected. Connected to non-master node. Aborting.')
+        sys.exit(9)
+
+
+def get_client(**kwargs):
```
Kami: Personally, I would prefer to explicitly define function signature instead of relying on **kwargs. It's easier to see what's going on.

packs/elastic/actions/search.py
```diff
@@ -0,0 +1,87 @@
...
+    def iselector(self):
+        """
+        Used to fetch indices/snapshots and apply filter to them.
+        """
+        if not self._iselector:
+            self._iselector = ItemsSelector(self.client, **self.config)
+        return self._iselector
+
+
+    def run(self, **kwargs):
```
Kami: Can you please explicitly declare method signature here? It looks like this class is only used by one action so it should be possible to do that without any other code changes.

packs/elastic/actions/search.py
```diff
@@ -0,0 +1,87 @@
...
+from collections import defaultdict
+import logging
+import sys
+import elasticsearch
+import json
+
+logger = logging.getLogger(__name__)
+
+
+class SearchRunner(ElasticAction):
```
Kami: Minor naming thing - I think `FindDocumentsAction` or similar is better since it's more obvious what the class / action does.

packs/elastic/actions/show.snapshots.yaml
```diff
@@ -0,0 +1,82 @@
...
+    default: days
+    description: Unit of time to reckon by [hours|days|weeks|months].
+    type: string
+  timeout:
+    default: 600
+    description: Elasticsearch operation timeout in seconds.
+    type: integer
+  timestring:
+    description: Python strftime string to match your snapshot's definition, e.g.
+      20150715020304 would be %Y%m%d%H%M%S.
```
Kami: Just curious - does this multi-line string works fine or it needs to be quoted?

packs/elastic/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+st2:
```
Kami: I don't remember if this is explicitly documented somewhere right now, but the convention we have been following so far is to put all the additional scripts and files in the `etc/` subdirectory.

packs/elastic/scripts/logsfeed.py
```diff
@@ -0,0 +1,99 @@
...
+import sys
+import os
+import datetime
+import time
+import random
+import socket
+from pprint import pprint
+from random import choice
+from jinja2 import Template
+
```
Kami: Short module level docstring on what this script does and how to use it would be nice :)

packs/elastic/scripts/seed.txt
```diff
@@ -0,0 +1,130 @@
+"Can this be still a dream?" Raskolnikov thought once more.
```
Kami: Just wanted to make sure the book this text is from is in a public domain?

Kami:
As the travis points out - pack is missing `pack.yaml` file.

Kami:
Since `elasticsearch` is the name of the product we are integrating with here and `elastic` is the name of the company, it would, imo, make sense and be better to call this pack `elasticsearch`.

Unless we also want to add actions for other products later on (but then we would still need to prefix the actions with product name, so something like `elastic.elasticsearch.indexes.optimize`).

What do you think?

Kami:
I think having "documents.insert" action would also be very useful (I already have many cool and useful use cases in mind myself - insert IRC messages into ES, insert Tweets into ES, etc.).

Users could then automatically insert information from other systems into ElasticSearch and we could write a blog post on how you can build poor man's (non-race safe) `X` (e.g. MySQL, PostgreSQL, etc.) -> ElasticSearch replication system using StackStorm (something like https://code.google.com/p/tungsten-replicator/ for example).

packs/elastic/README.md
dennybaa: I agree explicit way is good. I've taken the naming from curator which the pack is based on. Well this naming seems sane, *delete.**, *show.** - are explicit. Actions like,*open*, *alias*, *optimize* make sense only for indices, so I guess regrouping is not really needed or really helps with anything.

> This way actions which operate on the same resource are organized / grouped together ...
The are grouped by action operation rather than by the same resource. It's more likely an operation brings a similar interface for actions, rather than grouping by the resource (indices or snapshots)...

packs/elastic/actions/lib/static_metagen.py
dennybaa: It's not anything real so far, just a script for a custom case, but nice yeap)
If to develop this idea a bit more, than it makes sense documenting. For example, I could think of making it more generic plus include README.md generation for different parameters. Then it might be really handy.

packs/elastic/actions/lib/static_metagen.py
dennybaa: Thanks, shame on me)) Used to blocks in `ruby`)

packs/elastic/README.md
Kami: Yes, but `delete` can also operate on other resources (e.g. a document). 

And it's not just about explicitness, it's also about the consistency - all names having prefix vs some having prefix (on top of that, some other packs already follow this convention - `st2.kv.get`, `st2.actions.list`, etc.).

And using resource as a prefix means you can easily use prefix filter match to find the actions for a particular resource.

packs/elastic/actions/lib/utils.py
dennybaa: Just a reused peace of code, and I hijacked it with generator which chooses the proper parameters.
Unlikely, I didn't like the curator code(( it's pain to read and understand, well and if you see passing args like this it's totally inspired by curator.

packs/elastic/actions/lib/static_metagen.py
Kami: Ah, OK, I thought metadata files for curator actions were generated using this script.

packs/elastic/actions/lib/utils.py
Kami: So in this case we could improve on that and at least make it more readable in our pack :)

packs/elastic/actions/search.py
dennybaa: By two with different interfaces which are *search.q* and *search.body*...

packs/elastic/actions/show.snapshots.yaml
dennybaa: Surprisingly, it works fine)) even without quotes.

packs/elastic/docker-compose.yml
dennybaa: Oops, thanks got it.

packs/elastic/scripts/seed.txt
dennybaa: :+1:  )))

packs/elastic/actions/search.py
Kami: I see.

In this case it might make more sense to move common functionality into a base class and have two classes (one for each action) inherit from it - this way each class can implement `run` method with the respective arguments.

The whole idea of Python runner actions is that they are pretty much Python scripts / classes which can be re-used (or tested) outside of StackStorm action runner. Doing that `**kwargs` magic breaks this assumption since those scripts are now highly coupled to the metadata file and those scripts can't be easily re-used or tested without looking at the metadata file.

packs/elastic/README.md
dennybaa: Yeap so it will be *delete.document* or something with another entry point. Here grouping is made for curator stuff..
> ... some other packs already follow this convention - st2.kv.get, st2.actions.list, etc.).

A bit different, under `kv.*` for example they share the same interface, in case of elastic it happened in a different way.

> And using resource as a prefix means you ...

Yes, then it was worth making smth like *elastic.curator.delete.indices* - which looks too **bloated**, I guess.
If I want  cleaner way to invoke actions, it will make the way of finding particular one a bit more intricate.

If 4-level actions seem sane then it makes sense to group like you suggest. I totally agree with you, but all those you mention were 3-level actions. What do you think?

packs/elastic/actions/search.py
dennybaa: Hmm, got it interesting idea. I will uncouple than, though I don't prefer having different class for micro portions of code (in case with search). But following this idea it would be better to uncouple +1

packs/elastic/README.md
Kami: Yeah, I think 4 levels are too much, but I personally think "curator" shouldn't be included in the name.

This code using curator library is just an implementation detail. User shouldn't and doesn't care how it's implement - they just want to perform an action (e.g. optimize an index, delete a document, etc.). It doesn't matter to them how the action is implemented so having curator in the name, imo, doesn't make much sense and it adds confusion.

On the other hand, having a resource type on which the action operates on in the name does make sense imo - e.g. I want to find all the available actions on a document. If we follow the same naming convention, user can just search for actions starting with `elasticsearch.documents`.

jfryman: > This code using curator library is just an implementation detail. User shouldn't and doesn't care how it's implement - they just want to perform an action (e.g. optimize an index, delete a document, etc.). It doesn't matter to them how the action is implemented so having curator in the name, imo, doesn't make much sense and it adds confusion

This very much. :+1:. It's also what allows us to potentially change on a dime the backend to an action, assuming it has similar behavior/return contracts for users to consume.

packs/elastic/README.md
```diff
@@ -0,0 +1,170 @@
...
+These parameters include general options such as elasticsearch host, port etc.
+
+Parameter | Description | Default
+------------ | ------------ | ------------
+**host** | Specifies Elasticsearch host to connect to (**required**). | `none`
+**url_prefix** | Specifies Elasticsearch http url prefix. | `/`
+**port** | Specifies port remote Elasticsearch instance is running on. | `9200`
+**use_ssl** | Set to `true` to connect to Elasticsearch through SSL. | `false`
+**http_auth** | Colon separated string specifying HTTP Basic Authentication. | `none`
+**master_only** | Set to `true` enable operation only on elected master. If connected to a node other than master operation will fail. | `false`
```
jfryman: Would the library not automatically forward requests to the master?

packs/elastic/README.md
```diff
@@ -0,0 +1,170 @@
...
+"
+}
+```
+* Delete all indices matching *^logstash.\**:
+
+```
+st2 run elastic.delete.indices host=elk prefix=logstash
+```
+
+### Snapshot operations
```
jfryman: Is there a way to discover snapshots as well from an action, in the event I am not the person to issue the snapshot command?

jfryman: Ah, I see at https://github.com/StackStorm/st2contrib/pull/198/files#diff-3b3efd5a0011d6a49ca50ef50ab57ac2R23

packs/elastic/actions/lib/utils.py
jfryman: :+1:

packs/elastic/actions/alias.yaml
```diff
@@ -0,0 +1,88 @@
...
+  exclude:
+    description: Index list to exclude from the operation.
+    items:
+      type: string
+    type: array
+  host:
+    description: Elasticsearch host.
+    required: true
+    type: string
+  http_auth:
```
jfryman: I would split this out to `username/password`, if only because we are working on backend schema changes to prevent leakage of secrets. If there is a munged text field here, we'll have leakage for sure.

packs/elastic/actions/show.snapshots.yaml
```diff
@@ -0,0 +1,82 @@
...
+      type: string
+    type: array
+  host:
+    description: Elasticsearch host.
+    required: true
+    type: string
+  http_auth:
+    description: 'Use Basic Authentication ex: user:pass'
+    type: string
+  log_level:
```
jfryman: Maybe an `enum` type here then?

https://github.com/StackStorm/st2contrib/blob/b5c49e7b0dc2da0a9f29462c0e2adf80b443ea24/packs/jira/actions/create_issue.yaml#L15 for example.

jfryman:
This is super-:cool:. I added a few comments scattered about on metadata that should apply to all of them, and then echo @Kami's comments around namespacing. Keep all the action names close to their intent, not the library.

packs/elastic/README.md
dennybaa: No it doesn't forward anything, just exists with an error. I haven't reworked client connection implementation.

packs/elastic/README.md
dennybaa: Yeah. It's possible, *show.snapshots* then *delete.snapshots*, but only these two operate on snapshots.

The `elastic.snapshot` action actually operates on indices, you only give the repository name to store snapshot, and indices selection options, and as the result you'll get a **curator-%%%blabla_date_time** of **selected** indices.

jfryman: Got it. I'd make sure that anywhere you have manipulation commands, you also have introspection commands so the user doesn't have to leave stackstorm to get data.

packs/elastic/actions/show.snapshots.yaml
dennybaa: I'd pleased (it's nice to know) if you guys point me to a place where all the types are coded, since of course I couldn't find it docs. Maybe I'm as always too distracted and missed it:)

dennybaa:
@jfryman, yeah we discussed this with @Kami . First I'd like to make kwargs more clear where it's reasonable. And the second 
> Keep all the action names close to their intent, 

That's a good phrase) I'd like to try it. So you'd like to see:

- indices.delete
- indices.show
- indices.optimize

I guess you are both very right, that's my barely first python experience. And that hardcore implementation of curator influenced the way I have delivered the code. 
Moreover your suggests for these refactoring might  even simplify the current implementation, I'm starting to think:)

packs/elastic/README.md
dennybaa: With this current limited subset of commands this true - no need to leave to get the introspection. Plus dry_run is available, which is neat for *deletish stuff*...

jfryman: :cool:

dennybaa:
New PR for the updated pack https://github.com/StackStorm/st2contrib/pull/211

PR #199, opened 2015-05-13
Initial Email Integration

jfryman:
This PR aims to add initial support for email support. This includes:

* SMTP Sensor
* IMAP Sensor

## SMTP Sensor

The SMTP Sensor runs a local server as defined in `config.txt`, receiving email messages and emitting triggers into the system. This server is indiscriminate with what messages are received, and for whom. Basically, a catch-all for emails into the system.

As such, things like email filtering should happen upstream, or this should be run in a controlled environment.

## IMAP Sensor

The IMAP Sensor logs into any number of IMAP servers as defined in `config.txt`, polling for unread email messages and emitting triggers into the system. Sensor looks at one account and one folder at a time, and can be independently configured.


jfryman:
There is an error here: https://travis-ci.org/StackStorm/st2contrib/builds/62718108#L417-L418, has to do with the Monkeypatching necessary for getting `asyncore` to work with `eventlet`.

Otherwise, this is ready for some :eyes:.

packs/email/sensors/smtp_sensor.py
```diff
@@ -0,0 +1,124 @@
...
+        pass
+
+    def update_trigger(self, trigger):
+        pass
+
+    def remove_trigger(self, trigger):
+        pass
+
+
+class St2SMTPServer(smtpd.SMTPServer):
```
Kami: In think case it should be safe to ignore the pylint ignore and add "# pylint: disable=no-member" to the end of this line.

packs/email/sensors/smtp_sensor.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, localaddr, remoteaddr, sensor_service, logger, trigger):
+        smtpd.SMTPServer.__init__(self, localaddr, remoteaddr)
```
Kami: Same here.

jfryman:
:green_heart:

PR #186, opened 2015-04-21
[salt] execution modules

jcockhren:
### Context
Stackstorm is not installed on the Saltstack master but needs to execute [Salt Execution Modules](http://docs.saltstack.com/en/latest/ref/modules/all/index.html#all-salt-modules)

### Proposed Solution
This action allows for the execution of various [Salt Execution Modules](http://docs.saltstack.com/en/latest/ref/modules/all/index.html#all-salt-modules) through the [Salt NetAPI](http://docs.saltstack.com/en/latest/ref/netapi/all/index.html). 

#### Explicit Execution Module Actions Included
- [x] [archive](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.archive.html): gunzip, gzip, rar, tar, unrar, unzip, zip_
- [x] [cloud](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.cloud.html): 'action', 'create', 'destroy', 'network_create', 'profile_', 'virtual_interface_create', 'volume_attach', 'volume_create', 'volume_delete', 'volume_detach'
- [x] [cmdmod](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.cmdmod.html): 'run', 'run_chroot', 'script'
- [x] [cp](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.cp.html): 'get_file', 'get_url', 'push', 'push_dir'
- [x] [cron](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.cron.html): 'ls', 'rm_job', 'set_job', 'set_env', 'rm_env'
- [x] [data](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.data.html): 'cas', 'getval', 'update', 'dump'
- [x] [event](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.event.html): 'fire', 'fire_master', 'send'
- [x] [file](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.file.html): 'access', 'chgrp', 'chown', 'directory_exists', 'file_exists', 'find', 'manage_file', 'mkdir', 'remove', 'replace', 'search', 'symlink', 'touch', 'truncate'
- [x] [grains](http://docs.saltstack.com/en/2014.7/ref/modules/all/salt.modules.grains.html): 'append', 'delval', 'get', 'remove', 'setval'
[...]

PR #204, opened 2015-05-14
Hook up pylint

Kami:
This branch hooks up pylint - it's now automatically ran on each commit.

In addition to that, it fixes some bugs and issues discovered by pylint.


packs/salt/actions/local.py
```diff
@@ -14,7 +14,10 @@ def run(self, **kwargs):
             st2 run salt.client module=pkg.install \
                     kwargs='{"pkgs":["git","httpd"]}'
         '''
-        self.generate_package(cmd=cmd)
+        # TODO: This is broken, fix it. I temporary disabled it to avoid pylint
```
Kami: @jcockhren Can you please have a look?

The action appears broken, but I don't have enough context to fix it (what to do with args and kwargs parameters which are passed to the action) so I rather just commented it for now.

It would also be better if we get rid of "kwargs abuse" and explicitly declare arguments in the method signature.

Makefile
```diff
@@ -1,11 +1,24 @@
 ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
+# All components are prefixed by st2
+COMPONENTS := $(wildcard /tmp/st2/st2*)
```
lakshmi-kannan: /tmp?

lakshmi-kannan: NVM. Saw that you are cloning st2 to /tmp. 

So the reason you need to clone st2 is because pylint would complain about import errors otherwise?

Kami: Correct, without adding st2 components to the pack, pylint is not aware of the base `Action` and `Sensor` class signatures, etc.

packs/salt/actions/local.py
jcockhren: @Kami `kwargs` will be used for the explicit actions that will be defined in #186. I made edits to this exact line. I'll rebase, resolve and address the non-use of kwargs.

PR #206, opened 2015-05-14
Twitter

manasdk:
New actions

* direct_message
* follow

### DIFF
--- a/packs/twitter/actions/direct_message.yaml
+++ b/packs/twitter/actions/direct_message.yaml
@@ -5,9 +5,9 @@
   enabled: true
   entry_point: "direct_message.py"
   parameters:
-    user:
+    screen_name:
       type: "string"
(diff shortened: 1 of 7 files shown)


packs/twitter/actions/direct_message.yaml
```diff
@@ -0,0 +1,15 @@
+---
+  name: "direct_message"
+  runner_type: "run-python"
+  description: "Direct message a user."
+  enabled: true
+  entry_point: "direct_message.py"
+  parameters:
+    user:
```
DoriftoShoes: Probably want to not use runner.  Since it is a runner parameter for other runners (local, remote), it could be confusing to content writers/users.

manasdk: Good point dcd07663d5aa4acde8143fa21d7b0e096edd775b

PR #210, opened 2015-05-16
Ansible integration pack

armab:
### Actions
* [x] `ansible.command` - Run single [Ad-Hoc command](http://docs.ansible.com/intro_adhoc.html) eg. module (`ansible` wrapper)
* [x] `ansible.command_local` - Perform single ansible Ad-Hoc command locally (`ansible` executable)
* [x] `ansible.playbook` - Run [Ansible Playbook](http://docs.ansible.com/playbooks.html) (`ansible-playbook`)
* [x] `ansible.vault.encrypt` - Encrypt ansible data files (playbooks, vars, roles, etc) (`ansible-vault`)
* [x] `ansible.vault.decrypt` - Decrypt ansible data files (playbooks, vars, roles, etc) (`ansible-vault`)
* [x] `ansible.galaxy.install` - Install role from [Ansible Galaxy](http://docs.ansible.com/galaxy.html) - hub of [community developed roles](https://galaxy.ansible.com/) (`ansible-galaxy`)
* [x] `ansible.galaxy.remove` - Remove the role installed from Ansible Galaxy (`ansible-galaxy`)
* [x] `ansible.galaxy.list` - List installed from Ansible Galaxy roles (`ansible-galaxy`)

Actions parameters are built to be close as possible to original parameters of original ansible executables, so user could work with it easily without finding any inconsistencies.

----

Vagrant configuration which demonstrates st2 + ansible pack usage and performs several simple ansible actions on two VMs: `node1` and `node2` from `master` VM.
https://github.com/armab/st2-ansible-vagrant


armab:
@Kami @jfryman take a look when you're available.

jfryman:
@armab yes, sorry for the lack of response. The team has been up in Canada at OpenStack Summit this week, so computer access has been limited. Will look at first thing next week.

armab:
@jfryman  It's ok, I read Twitter :+1:

packs/ansible/README.md
```diff
@@ -0,0 +1,61 @@
...
+* `command` - Run single [Ad-Hoc command](http://docs.ansible.com/intro_adhoc.html). It has all the regular parameters of `ansible` executable.
+* `command_local` - Perform single ansible Ad-Hoc command (module) locally.
+* `playbook` - Action to run [Ansible Playbook](http://docs.ansible.com/playbooks.html) (`ansible-playbook` executable).
+* `vault.encrypt` - Encrypt ansible data files (playbooks, vars, roles, etc) with password (`vault` executable).
+* `vault.decrypt` - Decrypt ansible data files (playbooks, vars, roles, etc) with password (`vault` executable).
+* `galaxy.install` - Install role from [Ansible Galaxy](http://docs.ansible.com/galaxy.html) - hub of [community developed roles](https://galaxy.ansible.com/) (`ansible-galaxy`).
+* `galaxy.list` - List installed from Ansible Galaxy roles (`ansible-galaxy` executable).
+* `galaxy.remove` - Remove the role installed from Ansible Galaxy (`ansible-galaxy` executable).
+
+## Examples
```
Kami: Nice job with the readme :+1:

Kami:
It looks like all those actions are ran locally on the same server where the StackStorm action runner service is running.

If so and there is no reason to ever needing to run those actions on a remote host, you use Python runner (`run-python-script`) instead of a local one.

Kami:
Edit: I just saw the action metadata files.

It looks like you always use sudo because in some cases the action might need sudo privileges. I assume that is the reason why you went with the local runner instead of the Python one?

packs/ansible/README.md
```diff
@@ -0,0 +1,61 @@
...
+#### `ansible.command` examples
+```sh
+# run ansible command with optional verbose parameter
+st2 run ansible.command args='hostname -i' verbose=vv
```
Kami: Hm, isn't this example missing `hosts` parameter?

Or it will by default try to run it on all hosts?

packs/ansible/actions/command.yaml
```diff
@@ -0,0 +1,108 @@
...
+    type: string
+  timeout:
+    description: "Action timeout in seconds. Action will get killed if it doesn't finish in timeout seconds"
+    type: integer
+    default: 600
+  hosts:
+    description: "Hosts"
+    type: string
+    position: 0
+  args:
```
Kami: I believe `args` argument should be required for `command` and `command_local` action, since running those actions without any args doesn't make sense, right? Or am I missing something?

packs/ansible/actions/galaxy.install.yaml
```diff
@@ -0,0 +1,52 @@
...
+  cwd:
+    description: "Working directory where the command will be executed in"
+    type: string
+  action:
+    description: "Action to use"
+    type: string
+    position: 0
+    default: install
+    immutable: true
+  roles:
```
Kami: I believe this parameter should be required, right?

Kami: Actually, it looks like either `roles` or `role-file` argument should be provided, but not both of them.

If so, the action code should be updated throw if neither is provided (sadly we can't handle mutually exclusive dependencies inside the metadata file right now).

packs/ansible/actions/lib/shell.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    Decorator to Apply replacements in a list of command line arguments.
+
+    :param attribute: Class attribute name which stores replacement rules.
+    :type attribute: ``str``
+    :return:
+    :rtype: ``callable``
+    """
+    def _replace_args(f):
+        def _wrapper(self, *args):
```
Kami: Please also use `functools.wraps` here -http://pymotw.com/2/functools/#wraps

packs/ansible/actions/vault.decrypt.yaml
```diff
@@ -0,0 +1,44 @@
...
+    type: integer
+    default: 60
+    immutable: true
+  action:
+    description: "Action to use against playbook file(s)"
+    type: string
+    position: 0
+    default: decrypt
+    immutable: true
+  files:
```
Kami: It would probably also be good to add a warning / note here that this file will be modified (encrypted) in place.

Same goes for decrypt.

Kami:
I've tested all the actions and besides some minor things I have noted in the comments above, everything works and looks fine. :+1:

packs/ansible/README.md
armab: You're right, it's a mistake.
Correct version is: https://github.com/armab/st2-ansible-vagrant/blob/master/ansible.sh#L55

packs/ansible/actions/command.yaml
armab: No,
there is also `list-hosts`: 
`st2 run ansible.command hosts=all list-hosts=true`

And I didn't removed simple "standalone" parameters like `version` and `help` which could be useful during chatops sessions, for example to get quickly ansible version.

packs/ansible/actions/galaxy.install.yaml
armab: Since my implementation is just simple wrapper, this part should be handled by Ansible itself and it gives good enough message: 

Running with both roles and role-file: 
```sh
./ansible-galaxy install abc.de -r file.txt
st2 run ansible.galaxy.install roles=abc role-file=cde.txt
```
Output: 
```sh
status: failed
result:
{
    "failed": true,
    "stderr": "Executed command "/opt/stackstorm/virtualenvs/ansible/bin/ansible-galaxy --role-file=cde.txt install abc"
",
    "return_code": 1,
    "succeeded": false,
    "stdout": "Usage: ansible-galaxy install [options] [-r FILE | role_name(s)[,version] | scm+role_repo_url[,version] | tar_file(s)]

Options:
  -h, --help            show this help message and exit
  -i, --ignore-errors   Ignore errors and continue with the next specified
                        role.
  -n, --no-deps         Don't download roles listed as dependencies
  -r ROLE_FILE, --role-file=ROLE_FILE
                        A file containing a list of roles to be imported
  -p ROLES_PATH, --roles-path=ROLES_PATH
                        The path to the directory containing your roles. The
                        default is the roles_path configured in your
                        ansible.cfg file (/etc/ansible/roles if not
                        configured)
  -s API_SERVER, --server=API_SERVER
                        The API server destination
  -f, --force           Force overwriting an existing role

See 'ansible-galaxy <command> --help' for more information on a specific command.

- please specify a user/role name, or a roles file, but not both
"
}
```

armab: However when running st2 action without any optional parameters: 

```sh
st2 run ansible.galaxy.install

id: 556301bb9c9938107a5c7074
status: failed
result:
{
    "failed": true,
    "stderr": "Executed command "/opt/stackstorm/virtualenvs/ansible/bin/ansible-galaxy install None"
",
    "return_code": 1,
    "succeeded": false,
    "stdout": "- downloading role 'None', owned by
- sorry, None was not found on galaxy.ansible.com.
- you can use --ignore-errors to skip failed roles.
"
}
```

So `st2 run ansible.galaxy.install` is transformed into `ansible-galaxy install None` which I don't think good behavior from st2 engine. Positional parameter still could be optional.

armab: Ok, I dirty-fixed it by adding `default` in yaml: 

```yaml
parameters:
  ...
  roles:
    description: "Role(s) to remove (separated by space)"
    type: string
    position: 1
    default: "" #  Added, removes `None` when this positional parameter is not set
```

so `st2 run ansible.galaxy.install` is transformed correctly into `ansible-galaxy install` and the output from ansible is: 

```sh
st2 run ansible.galaxy.install

...
- you must specify a user/role name or a roles file
```

packs/ansible/actions/lib/shell.py
armab: Thanks, nice hack!

packs/ansible/actions/vault.decrypt.yaml
armab: Added

armab:
About using `local-shell-script` instead of `python-script` runner. Good question.

At start I've played a bit with python-script runner and worked directly with [Ansible API](http://docs.ansible.com/developing_api.html). But after all there are many reasons why I choosed local-shell-script and that wrapper way: 

##### The first thing is why I didn't choosed Ansible Python API:
  * it wasn't that simple case when you redirect all parameters into API classes/functions and just get the results
  * a lot of conditional things should be copy-pasted from Ansible codebase to reimplement already existing behavior
  * that low-level behaviors still could be changed in Ansible codebase during time, while high-level usage (via command/parameters) may remain unchanged
  * in API all parameter names differs noticeably from cmd params, some parameter values could be changed automatically depending on union of other parameters
  * with a lot of optional parameters it becomes hell
  * Ansible API could be good if you know exactly what you need or when you need some focused/specific needs. For general usage it wasn't the way.
  * So after all, at this point of requirements I couldn't find any advantages of using Ansible API. There wasn't any case when I needed to do something customized (for example reformat output).

armab:
##### Why I choosed `local-shell-script` instead of `python-script` runner: 
* `local-shell-script` has `cwd` by default, which is useful in cases like:
    ```sh
    st2 run ansible.vault.decrypt cwd=/etc/ansible vault-password-file=vault.txt files='playbooks/1.yaml playbooks/2.yaml 3.yaml'
    ```

    Don't need it to implement it with `python-script`.
* In Yaml I have parameters with dashes and for consistency reasons I wanted to keep naming the same as it is in Ansible cmd, so with `python-script` this becomes less beautiful: 

    ```py
    def run(self, files, cwd, debug, **kwargs):
        password = kwargs.get('vault-password-file')
        ...
    ```

* When running Ansible from cmd, some of the parameters allow wildcards like: 
    ```sh
    st2 run ansible.vault.decrypt vault-password-file=vault.txt files=playbooks/*
    ```
    and this should be reimplemented on my side in `python-script` way.

* I also tried to launch the ansible executable from `python-script`, but found it impossible to build the cmd: 
    ```py
    def run(self, *args, **kwargs):
        # which one from kwargs is positional argument/which one is named? They all are unordered also.
    ```

----

Considering all reasons, keeping in mind that I want 100% the same behavior as Ansible has, the same parameter naming as Ansible has - I found `local-shell-script` wrapper around executable files ideal for this special case.

armab:
Also, don't know how rare this case is,
but during my work I found useful possibility to define some (pre/post) hooks inside the pack which would be executed on installation, on virtualenv create maybe or maybe before_first_run/after_first_run.
For example to create some directories/files used by pack and do clean-up on pack uninstall/unload.

In my special case for better user experience I wanted to set default `cwd` to `/etc/ansible`, which is hardcoded default Ansible directory. The problem is that `/etc/ansible` dir doesn't exist and thereby I can't set it as default cwd.

packs/ansible/actions/command.yaml
Kami: I see, thanks for the clarification.

packs/ansible/actions/galaxy.install.yaml
Kami: Ideally, we would know the required parameters before the run time, just by introspection the metadata (this way we can provide better validation error messages, generate more user-friendly UI forms, etc.), but we don't allow users to specify mutually exclusive parameters in the metadata file right now so your approach is fine in this case.

armab: Yes, I got the idea, it's Yaml side.

Kami:
Just FYI, build failure doesn't seem to be related to your change. It seems to be timing out because the whole build process takes too long.

I will look into fixing this in a separate PR later today.

packs/ansible/actions/lib/ansible_base.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Get full path to executable binary located in pack virtualenv.
+
+        :return: Full path to executable binary.
+        :rtype: ``str``
+        """
+        if not self.BINARY_NAME:
+            sys.stderr.write('Ansible binary file name was not specified')
+            sys.exit(1)
+        path = os.path.join('/opt/stackstorm/virtualenvs/ansible/bin', self.BINARY_NAME)
```
armab: The only bad thing related to `local-shell-script` is hardcoding the path ^^ here.

I found that with `python-script` it's possible to get virtualenv path in a good way: 
```py
st2common.util.sandboxing.get_sandbox_virtualenv_path(pack='ansible')
```

Kami:
Merged into master. Thanks.

packs/ansible/actions/lib/ansible_base.py
Kami: Yep.

armab:
Thanks for discussion!

packs/ansible/actions/lib/ansible_base.py
jfryman: Think it'd be too terribly difficult to expose this as an ENV variable for shell runners? I could see this (and that pattern) coming in handy

jfryman:
This is a great PR! Thanks so much for the code!

PR #211, opened 2015-05-18
Elastic updates

dennybaa:
Updated elastic pack.
* Problems which implicit parameters were solved where it makes it reasonable.
* Readme updates.
* Logstash integration tips.


Kami:
Thanks. I will have a look shortly.

packs/elastic/actions/search.py
```diff
@@ -0,0 +1,89 @@
+from easydict import EasyDict
+from lib.items_selector import ItemsSelector
+from lib.esbase_action import ESBaseAction
+from collections import defaultdict
+import logging
+import sys
+import elasticsearch
+import json
+
+logger = logging.getLogger(__name__)
```
Kami: You can use `self.logger` attribute which is available on the action classes.

packs/elastic/scripts/logsfeed.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
```
Kami: Comments from the previous PR still apply - please move additional scripts and files into `etc/` directory.

packs/elastic/pack.yaml
```diff
@@ -0,0 +1,10 @@
+---
+name : elastic
```
Kami: I think "elasticsearch" is more appropriate name for this pack. See comment from the previous PR - https://github.com/StackStorm/st2contrib/pull/198#issuecomment-101238465

Kami:
There are some pep8 and pylint failures which need to be fixed - https://travis-ci.org/StackStorm/st2contrib/builds/63086792

packs/elastic/pack.yaml
Kami: And other comments which haven't been addressed yet from the old PR (https://github.com/StackStorm/st2contrib/pull/198) still apply :)

packs/elastic/actions/search.py
dennybaa: I'd love to. Definitely logging should be reworked. But I don't know how yet...
There are some problems with curator, if it's possible to route it to use the same logger which is self.logger it would be cool.
Plus there are whitelists (or black) I don't really remember, which are also used in curator.

packs/elastic/pack.yaml
dennybaa: We've discussed kwargs with manas, if you mean them. He thinks going with kwargs is just fine. Having as many classes as required in such a case will just an overkill.
Btw `get_client` has been fixed to use explicit parameters.

dennybaa: > I think "elasticsearch" ...
That's right. Though we can use this pack for other elastic products like logstash and kibana. What do you think?

Kami: I'd rather prefer to have a pack per product / tool, since in this case, those products don't have much in common besides the parent organization.

I don't think we could re-use any code and this means we would also need to prefix actions with the service name - e.g. `elasticsearch.indices.optimize`.

dennybaa: k. I guess I need to rename it then.

packs/elastic/scripts/logsfeed.py
dennybaa: Ok I'll move them, docker-compose.yml as well?

It's not a handy convention for `docker-compose.yaml`. Since if it's in 
the root of `elasticsearch/`it will be nice to build, run and exec 
because containers get names like **elasticsearch_st2_1** and 
**elasticsearch_elk_1**.

For scripts it would be nice, but for `docker-compose.yaml` I would 
prefer to leave it in its place.

Kami: I'm slightly confused - what changes if we move it to etc/ directory?

In any case, not moving that file is not a blocker right now so if it makes things easier, do leave it there.

On a related note - similar as other scripts, docker-compose.yaml is also not really useful without docs. It would be great to add some docs on how to use it to the readme.

dennybaa: Moving files it's ok no problem. But for the compose file I will it in 
its place, because the default naming is nice.

It's just for testing purpose for dev usage. When you test a pack and 
bringing up docker env using compose you just know right away into 
which container to login)

dennybaa:
Cool stuff this pylint. I've fixed `chunk_index_list` invocation.
Can I somehow add scenario for pylint to bypass  errors like these:
```
E:  8, 9: Module 'logging' has no 'getLogger' member (no-member)
E:121,15: Instance of 'EasyDict' has no 'timestring' member (no-member)
```

Kami:
Yeah, you can add "disable" pragma to the end of the line in questions - `# pylint: disable=no-member`.

In general, you should also always try to figure out the root cause first before simply ignoring an error - in a lot of cases, it detects actual issues, but in some cases, it detects false positives (especially if there is a lot of meta programming and "magical" code).

PR #212, opened 2015-05-21
Update st2 command aliases

Kami:


packs/st2/aliases/actions_list.yaml
```diff
@@ -1,4 +1,7 @@
 ---
-name: "st2-actions-list"
+name: "st2_actions_list"
 action_ref: "st2.actions.list"
 description: "List available StackStorm actions."
+formats:
+    - "list actions"
```
Kami: I think we should probably still prefix all those commands with "st2" (e.g. `!st2 list actions`). 

What do other ppl think?

jfryman: I do prefer that.

manasdk: Yes prefix `st2`

manasdk:
+1

PR #213, opened 2015-05-27
[STORM-1344] Replace stormbot_say with hubot.post_message

manasdk:
* using the hubot pack instead of stormbot_say

### DIFF
--- a/packs/nagios/rules/nagios_service_chat.yaml
+++ b/packs/nagios/rules/nagios_service_chat.yaml
@@ -9,7 +9,8 @@ criteria:
     pattern: 2
     type: gt
 action:
-  ref: stormbot_say
+  ref: hubot.post_message
   parameters:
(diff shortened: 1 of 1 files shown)


Kami:
LGTM, +1

jfryman:
This is a simple enough change, but here is a concern of mine.

What if I don't have Hubot? What if my ChatOps adapter is something different? I've been trying to think about an adapter of convention for things like this that don't lock us to something specific. I think this is going to come up often.

manasdk:
> What if I don't have Hubot?

The odd part here is that we are trying to make rules generally available. Rules are often infrastructure/install base specific so this rule should probably not even exist and shared publicly with an expectation that it would actually work.

To answer you original question I think an adapter approach is required. We could expect that users would setup `chat.post_message` action and back it with their preferred impl. Similarly, we could provide other canonical actions that can come with implementations that are infra specific e.g. `server.create`, `ticket.create` or `page.acknowledge` with each of them being backed by a specific action.

Anyway, for now I think I will merge this change as stormbot_say has been removed and to maintain this rule we should use the hubot pack.

PR #214, opened 2015-05-27
Add scrutinizer.com badge

armab:
I saw there is a lot of work done regarding `flake8`, `pylint` and other static code analysis checks with Travis.

This is proposal to add scrutinizer (which uses pylint too) and include that another one "opensource-welcome" badge :)
[Scrutinizer Code Quality](https://scrutinizer-ci.com/g/armab/st2contrib/?branch=master)

I've polished Scrutinizer config synchronizing it with [pylint](https://github.com/StackStorm/st2contrib/blob/master/.pylintrc) disabled checks, so Scrutinizer shows only meaningful issues.

For example I liked [this bug warning](https://scrutinizer-ci.com/g/armab/st2contrib/issues/master/files/packs/salt/actions/lib/base.py?orderField=path&order=asc#inspectioncomment-856251098), not sure why it didn't showed with `make pylint`, maybe Scrutinizer has some unique sniffs.
 
More: https://scrutinizer-ci.com/g/armab/st2contrib/issues/master

-----

PS. To enable this, just don't forget to add repo in https://scrutinizer-ci.com/

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -16,9 +16,9 @@ build:
             #- for f in packs/*/requirements.txt; do pip install -r $f; done
 
 # https://scrutinizer-ci.com/docs/configuration/build_status
-build_failure_conditions:
-    - 'elements.rating(<= D).new.exists'     # No new classes/methods with a rating of D or worse
-    - 'issues.severity(>= MAJOR).new.exists' # No new issues of major or higher severity
+#build_failure_conditions:
+#    - 'elements.rating(<= D).new.exists'     # No new classes/methods with a rating of D or worse
+#    - 'issues.severity(>= MAJOR).new.exists' # No new issues of major or higher severity
 
 checks:
     python:
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,163 @@
+build:
+    environment:
+        python: 2.7.7
+        # disable uneeded software
+        mysql: false
+        postgresql: false
+        redis: false
+    dependencies:
+        before:
+            ## still false-positives after installing dependencies
+            ## just disable 'imports_import_error' check
+            #- sudo apt-get install -y software-properties-common
+            #- sudo add-apt-repository 'deb http://downloads.stackstorm.net/deb/ trusty_unstable main'
+            #- sudo apt-get update
+            #- sudo apt-get install -y --force-yes st2reactor st2common st2auth st2api st2actions st2debug
+            #- for f in packs/*/requirements.txt; do pip install -r $f; done
+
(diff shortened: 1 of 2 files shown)


manasdk:
does not use .flake8 and .pylintrc?

armab:
Yes and No :)

They have [Pylint check which reads .pylintrc](https://scrutinizer-ci.com/docs/tools/), but it's kind of separated check and config is not merged with their own named parameters.

So `.pylintrc` could be used: https://scrutinizer-ci.com/docs/tools/python/pylint/ for some results, but to get more custom sniffs still need to tick/test those named parameters in their UI.

armab:
Also [CodeClimate](https://codeclimate.com/) introduced Python checks 1 month ago maybe, in beta for the moment: 
 https://twitter.com/codeclimate/status/587614231681310721

But they don't have anything useful for now: https://codeclimate.com/github/armab/st2contrib/issues

lakshmi-kannan:
I have some personal gripes about using something like scrutinizer. We already have linter, flake8, pylint etc. This adds more friction to committing code without much perceivable benefits. What values does scrutinizer add that is not available from flake8, pylint etc? Maybe I am missing something.

armab:
* they have some custom code checks
* it's possible to hide specific false-positive via web interface, instead of suppressing entire rule like this: https://github.com/StackStorm/st2contrib/blob/master/.pylintrc#L10
* it shows extended things on classes/modules like Duplication/Size/Complexity/Rating in easy web interface: https://scrutinizer-ci.com/g/armab/st2contrib/code-structure/master
* it shows repository statistics/improvements over time: https://scrutinizer-ci.com/g/armab/st2contrib/statistics/
* it doesn't seems it would bring some overhead when committing new code
* it becomes clear that repo follows some usual coding standards just from the first sight
* badge with rating

It looks like those badges doesn't matters, but I believe that people more likely to pay attention when they see such signals such as code coverage/build/license in graphic form, visual perception is very effective.
So in my understanding the biggest benefit here is showing "code quality" flag and attitude to community that repo is alive, active and friendly.

armab:
Ok,
I just disabled build failure conditions for Scrutinizer, so the build won't be Red and never annoy anyone.

This allows Scrutinizer to be just optional with their custom checks, web interface and code quality rating.

lakshmi-kannan:
@armab: As long as these are suggestive and are not going to break the build, I am +1 to merging this.

Thanks for pointing out scrutinizer metrics. Indeed the visual metrics are great.

Kami:
Yeah, I'm also fine with that as long as the metrics are suggestive and don't block / break the build.

armab:
Cool! :tada:

PR #215, opened 2015-05-27
chef pack update, chef.apply action introduced

dennybaa:


dennybaa:
This is more interesting) The linting lints all the code. In my branch there's even no elasticsearch at all...
This seems to be an issue. It should not lint that code which is not in my brunch, I guess.
Is it possible and reasonable to setup travis like that?

Moreover, elasticsearch has already passed linting issues https://travis-ci.org/StackStorm/st2contrib/builds/63546306, but they are fired again.

With the setup like this it's possible to get unpredictable issues. When all of a sudden config is changed and there's already no possibility to merge your PR unless fixing "some other code".

dennybaa:
It would be nice if the code which is being merged is only linted... It will also speed up the whole process.

dennybaa:
@Kami , you might have ideas why such things happen. I've looked at the code, in my sublime with pylint on there's nothing. Maybe because of that that curator is installed on my machine, and it's possible to resolve the name, what do you think?
https://gyazo.com/93084d486ed70aac6845ebb21d4b4e27

And what's also a bit strange is that pylint doesn't show the exact ID of error... 
```
E: 86,20: Undefined variable 'get_snapshots' (undefined-variable)
E:102,18: Undefined variable 'get_indices' (undefined-variable)
```

Kami:
@dennybaa Yeah, that's correct - dependency is not available so it can't detect members from it.

A while ago I added support for creating a virtualenv and installing all the dependencies for each pack, but it took too long and started timing out a while ago so sadly I needed to revert that change - fb90ebc3650b5dad333faa8578b061fa70cd6e54.

dennybaa:
@Kami virtual env and installation it's obviously not an option.

@Kami , but what about what I've said. Is it possible to check the code which is exactly being merged? Only those specific files, not the whole repo.

Kami:
@dennybaa I don't think you can do it directly with Travis (we could do it in our wrapper script though), but I still prefer to run pylint on every file.

This way it's "idempotent" and always fails if there is a bad file somewhere. If we only linted changed files, it's possible that we would end up with bad files in the repo without noticing it if someone (accidentally) pushed or merged bad files in the repo.

Kami:
One thing I will try is to only create a single virtualenv for all the packs - this should work fine as long as two packs don't have conflicting dependencies (e.g. same package but a different version).

dennybaa:
In this case, I see the only reasonable solution is to have per pack
repository. No conflicts, also virtual env dependencies can be fetched.

dennybaa:
One virtual env for all the packs it's not a good idea, just a workaround,
also with unpredictable results. The performance of such approach will be
also bad.

Kami:
@dennybaa One virtualenv for all the packs should work fine right now since there are no conflicting dependencies as of now.

As far as pack per repository - we already kinda support that (installing packs from a user-specified repo), but splitting out an official st2contrib repository is something which has already been planned for a while now, but we haven't had the time to do it yet.

To split it out we still need to figure out, decide and implement some things (e.g. how are we going to handle client and web discovery - are we just going to use git subtrees, pack versioning, pining pack version to a specific StackStorm version, etc.).

I think we will do that once we work on implementing the pack management API - so probably some time in the next couple of months.

Kami:
@dennybaa I've merged #216. Can you please sync up your branch with latest master and make sure it resolves the issue?

packs/chef/actions/chef_client.py
```diff
@@ -8,14 +8,14 @@ class ChefClientRunner(ChefRunner):
     ChefRunner type implementation.
     Invokes chef-client binary with given arguments.
     '''
-
-    chef_binary = 'chef-client'
```
Kami: Any particular reason why you moved this variable?

packs/chef/actions/chef_install.py
```diff
@@ -20,9 +19,10 @@ class ChefInstaller(object):
         ('-d', '--download_path', {})
     ]
 
-    def install(self, install_method, options):
+    @staticmethod
+    def install(install_method, options):
         exit_code = 0
-        installer = getattr(ChefInstaller, install_method.capitalize())(options)
+        installer = getattr(sys.modules['lib'], install_method.capitalize())(options)
```
Kami: Hm, why do we need to do that?

packs/chef/actions/apply.yaml
```diff
@@ -0,0 +1,29 @@
...
+    execute:
+        description: "Execute resources supplied in a string"
+        type: string
+    minimal_ohai:
+        description: "Only run the bare minimum ohai plugins chef needs to function"
+        type: boolean
+    recipe_file:
+        description: "Path to the recipe file to be executed (relative to a remote host)"
+        type: string
+    timeout:
```
Kami: `timeout` is already a remote runner parameter which means you don't (and shouldn't) specify it in the action metadata file unless you want to change the default value and / or make it immutable.

To confirm it - I assume you specified it because you want to increase the default timeout from `60` to `300` seconds?

dennybaa: Yeap, of course, and also might be description clarification (later was in elasticsearch pack).

packs/chef/actions/chef_install.py
dennybaa: Yeap, *because of linting*. `install_method` comes from `__all__` and it happened that linting wasn't happy and didn't see the attribute.

packs/chef/actions/chef_client.py
dennybaa: Not, sure why) There are different values for this variable now. It couldn't be a reason though, just decided that it's more neat.

packs/chef/actions/chef_install.py
Kami: Looking at the code - it doesn't look like the previous code ever worked since the " from omnibus .." line is commented.

Kami: In any case, I would prefer to have an explicit mapping instead of doing this unnecessary import hackery which makes code hard to understand.

We only support one installation method anyway so this is really an over-kill.

Something like this is better:

```python
INSTALLER_NAME_TO_CLASS_MAP = {
  'omnibus': 'OmniBus
}

...
installer = dict.get
....
```
And yeah, I know you didn't originally contribute this code, but if we are already touching the code, let's improve it :)

Kami: And I imagine that just changing the code to:

```python
installer = getattr(self, install_method.capitalize())(options)
```
and commenting out the import inside the class would work as well.

dennybaa: Yeap. Both ways are nice, so I will change it the second one. Let's do the second way, interesting) I hope linting will be okay about such import.

dennybaa: Okay, the second way was linted successfully on my node. Waiting for 
travis. Now with all pips it's slow again(

packs/chef/actions/chef_client.py
Kami: I personally prefer class variable (previous) approach since it's more clear in this case and in this case it also works fine.

It's also worth noting that class variable is not exactly the same as instance variable, but in this case either one works fine.

Kami:
Besides the class variable thing it looks good to me.

(I still need to test it end to end though :))

dennybaa:
Sure, if you wan't to test it I suggest `docker-compose`. Just build recent st2 container with st2express, fix `docker-compose.yaml` (if needed) and you are ready to go.

dennybaa:
Also I'd like to vote for testkitchen.ci, if when we have st2 containers on dockerhub (i guess this might be already in progress), kitchen docker works really smooth. It will really easy to introduce integration testing for packs like this (chef, salt, ansible etc) and for sure other packs also :)

jfryman:
@dennybaa I know the team is looking at testkitchen for our end-to-end tests. Likewise, we just published Docker images this weekend if you want to help us test them before we go publicize a ton. https://registry.hub.docker.com/repos/stackstorm/ and http://docs.stackstorm.com/latest/install/docker.html#distributed-containers

Kami:
@dennybaa If I specify `minimal_ohai` parameter the action fails.

```
id: 558a6bc20640fd28a9f522d7
status: failed
result: 
{
    "localhost": {
        "failed": true, 
        "stderr": "/usr/lib/ruby/vendor_ruby/mixlib/cli.rb:229:in `parse_options': invalid option: --minimal-ohai (OptionParser::InvalidOption)
	from /usr/lib/ruby/vendor_ruby/chef/application/apply.rb:82:in `reconfigure'
	from /usr/lib/ruby/vendor_ruby/chef/application/apply.rb:158:in `run'
	from /usr/bin/chef-apply:25:in `<main>'", 
        "return_code": 1, 
        "succeeded": false, 
        "stdout": ""
    }
}
```

Is this option only available in the latest version or is something else going on?

Kami:
Answering my one question above - yes, looks like this option is only available in chef-client >= 12.3 (https://docs.chef.io/release_notes.html).

I will add a note to the README.

Kami:
Tested the changes, they work fine so I went ahead and merged PR intro master.

Thanks.

dennybaa:
Cool! You've found it, it's appeared in chef 12 right.

PR #217, opened 2015-06-01
Add "disable_formatting" parameter to the slack.post_message action

Kami:
With this parameter user can disable formatting and parsing of the message.

screenshot from 2015-06-01 16 46 44

### DIFF
--- a/packs/slack/actions/post_message.py
+++ b/packs/slack/actions/post_message.py
@@ -11,7 +11,8 @@ __all__ = [
 
 
 class PostMessageAction(Action):
-    def run(self, message, username=None, icon_emoji=None, channel=None):
+    def run(self, message, username=None, icon_emoji=None, channel=None,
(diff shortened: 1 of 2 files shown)


jfryman:
:+1:

manasdk:
Do we get support for this in hubot-stackstorm as well?

jfryman:
@manasdk good question. I think that needs to be there as well.

Kami:
Yes - for that, we need to update hubot-slack adapter and JavaScript slack client.

I have opened PRs against respective the projects. If the PRs won't get merged soon, we will sadly need to resort to forking those projects.

jfryman:
@manasdk actually, I don't think that matters... right as I hit enter. hubot-stackstorm is already receiving the raw end from StackStorm, but the client may still render which is why there is the PR @Kami posted earlier

PR #218, opened 2015-06-01
travis-ci integration pack

Aamir-raza-1:


packs/travisci/config.yaml
```diff
@@ -0,0 +1,7 @@
+git_auth_key: 7c2decb3962a2bfa2a5351d3614004cfeee3f92e
```
jfryman: Might not want to commit these things. :)

packs/travisci/actions/hook.yaml
```diff
@@ -0,0 +1,15 @@
+name: hook
```
jfryman: Try making this a more descriptive action name. Like `setup_hook`.

packs/travisci/actions/hook.yaml
```diff
@@ -0,0 +1,15 @@
...
+runner_type: run-python
+description: enable or disable hooks for git repos, defaults to false if active argument not passed
```
jfryman: I would not add a default. You can make the parameter required, but not provide a default. That way, user needs to be explicit in executing action, and cannot accidentally cause a problem.

packs/travisci/actions/list_repos.py
```diff
@@ -0,0 +1,19 @@
+from lib.action import TravisCI
+import requests
```
jfryman: I didn't see this lib in the `requirements.txt`. I know we ship this by default (https://github.com/StackStorm/st2/blob/master/requirements.txt#L17), but it's good to be explicit so the library is setup in the pack virtualenv in the event we decide to change libraries.

packs/travisci/actions/build_repo.yaml
```diff
@@ -0,0 +1,14 @@
+name: build_repo
```
jfryman: Can I pass an arbitrary git `ref` to this action? It's good to be able to kick off builds against `master`, but equally beneficial to be able to kick off builds against any branch.

jfryman:
Also want to add a `README.md` to this pack.

jfryman:
:cool: Added several comments. Good start.

packs/travisci/actions/hook.yaml
Aamir-raza-1: Ok, I have noted the comments, I'll improve these and then push up the stream. Thanks

packs/travisci/actions/list_repos.py
Aamir-raza-1: Sure, I will add this to requirements.txt

packs/travisci/actions/build_repo.yaml
Aamir-raza-1: Yeah, I got your point, will improve that way

Kami:
Sorry for the delay. I will review and test the changes today.

packs/travisci/config.yaml
```diff
@@ -0,0 +1,7 @@
+git_auth_key: Your Git Authentication Key
```
Kami: We should simplify the config - only thing which user really needs to specify is the Travis Auth token.

Kami: Also, it currently looks like `git_auth_key` configuration value is not used anywhere?

packs/travisci/actions/list_repos.py
```diff
@@ -0,0 +1,19 @@
...
+import yaml
+
+
+class GetRepos(TravisCI):
+    def run(self, username):
+        """
+        Listing all Repos for a give user
+        """
+        _HEADERS = self.travis
```
Kami: You should be able to reduce the amount of duplicated code by refactoring this functionality into a new `_perform_request(url, query_params, method='GET')` or a similar method on the base action class.

This method should perform the request and return the parsed response.

packs/travisci/actions/list_repos.py
```diff
@@ -0,0 +1,19 @@
...
+        _HEADERS['Authorization'] = self.config["Authorization"]
+        uri = self.config["uri"] + '/repos?owner_name=' + username
+        response = requests.get(uri, headers=_HEADERS)
+        data = yaml.load(response.content)
```
Kami: I'm slightly confused, we send "Accept: application/vnd.travis-ci.2+json" header, but you are using YAML to parse the response here?

packs/travisci/actions/get_branches.py
```diff
@@ -0,0 +1,18 @@
+from lib.action import TravisCI
+import requests
+import yaml
+
+
+class GetBranches(TravisCI):
```
Kami: Minor thing - for consistency reasons, every action class should have `Action` suffix, so `ListBranchesAction`.

It's also better to use `List` instead of `Get`. We usually use get when you retrieve a single instance and list when you retrieve more.

packs/travisci/actions/get_builds.py
```diff
@@ -0,0 +1,25 @@
+from lib.action import TravisCI
+import requests
+import yaml
+
+
+class GetBuilds(TravisCI):
```
Kami: ListBuildsAction :)

packs/travisci/actions/get_hooks.py
```diff
@@ -0,0 +1,20 @@
+from lib.action import TravisCI
+import requests
+import yaml
+
+
+class GetHooks(TravisCI):
```
Kami: ListHooks

packs/travisci/actions/setup_build.yaml
```diff
@@ -0,0 +1,14 @@
+name: setup_build
```
Kami: It would probably be better and more explicit to have two actions - `restart_build` and `cancel_build`.

packs/travisci/actions/setup_hook.yaml
```diff
@@ -0,0 +1,14 @@
+name: setup_hook
```
Kami: Same here - probably better to have two actions - `enable_hook_for_repo`, `disable_hook_for_repo`.

packs/travisci/requirements.txt
```diff
@@ -0,0 +1 @@
+requests
```
Kami: We should specify a min version, e.g. `requests>=2.6.0`.

Kami:
@Aamir-raza-1 I've added some comments. Once those are addressed we can get this PR merged.

Thanks.

Aamir-raza-1:
@Kami, Ok sure I would fix these.

Kami:
@Aamir-raza-1 Thanks.

Please ping me when they are addressed and I will have a look.

Aamir-raza-1:
Yea sure, I will.

Aamir-raza-1:
@Kami Please review the changes that i made.

Kami:
@Aamir-raza-1 Thanks, I will look over it today, make some minor style changes and fixes and get it merged.

Kami:
I spent more than an hour reviewing and fixing the PR and I'm still not done yet (I lost the motivation). I've opened a new PR which includes work in progress cleanup and fixes - #234.

It looks like you either missed or ignored a bunch of the comments I added a while ago (simplifying the config, using list instead of get, ...). In addition to that a bunch of actions were broken and didn't work (see my commits for which ones).

On a related note - we also need better documentation on retrieving the Travis CI API access token. The API docs (http://docs.travis-ci.com/api/#authentication clearly state that access token != Travis CI token (` Do not confuse the access token with the token found on your profile page.`) which is not clear from the docs right now.

Can you please be more careful and pay more attention to details next time?

Thanks.

Aamir-raza-1:
Ok, sure, I tested all actions before pushing them upstream, but I will definitely now again pull the changes and test them. I guess something minor would have been missed out while you improved the code. I will look into this and will let you know. Thanks

PR #223, opened 2015-06-07
Resolving: post message returns status 'success', even though it fails

kirilstrax:
I am trying to run a slack.post_message and it fails with error, and goes into 
    else:
            self.logger.exception('Failed to post message: %s' % (response.text))
branch. The overall return status is, however, still 'Succeeded'. So suggesting to raise the received message in this case as exception:
    raise Exception(response.text)


Kami:
Fixed lint issue and merged change into master. Thanks.

PR #225, opened 2015-06-10
[IMAP Sensor] Support for automatically downloading and storing message attachments in the datastore

Kami:
This pull request adds support for automatically downloading and store message attachments in the datastore.

This allows users to retrieve those attachments from a datastore later (e.g. inside an action).

For more information on how it works, pleasae refer to the readme.

### DIFF
--- a/packs/email/README.md
+++ b/packs/email/README.md
@@ -45,10 +45,10 @@ The following attachment settings can be configured:
 
 * ``download_attachments`` - True to download the attachment and store them in the
   datastore.
-* ``max_attachment_size`` - Maximum size of download attachment. If an attachment
-  excedes this size it will be skipped and not downloaded.
-* ``attachment_datastore_ttl`` - TTL for the attachment value which is stored in the
-  datastore.
+* ``max_attachment_size`` - Maximum size of download attachment bytes. If an
+  attachment exceeds this size it will be skipped and not downloaded.
+* ``attachment_datastore_ttl`` - TTL in seconds for the attachment value which is
+  stored in the datastore.
 
 If ``download_attachments`` attribute for a particular IMAP server is set to ``True``,
 attachments will be automatically downloaded and stored in the built-in datastore under
--- a/packs/email/README.md
+++ b/packs/email/README.md
@@ -28,6 +28,7 @@ imap_mailboxes:
     username: "james@stackstorm.com"
     password: "superawesomepassword"
     ssl: True
+    download_attachments: False
   alternate:
     server: "mail.stackstorm.net"
     port: 143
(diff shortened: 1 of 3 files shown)


packs/email/README.md
```diff
@@ -35,4 +36,150 @@ imap_mailboxes:
...
+    download_attachments: True
+max_attachment_size: 1024
+attachment_datastore_ttl: 1800
+```
+
+The following attachment settings can be configured:
+
+* ``download_attachments`` - True to download the attachment and store them in the
+  datastore.
+* ``max_attachment_size`` - Maximum size of download attachment. If an attachment
```
manasdk: unit?

packs/email/sensors/imap_sensor.py
```diff
@@ -101,7 +135,57 @@ def _process_message(self, uid, mailbox):
...
+
         self._sensor_service.dispatch(trigger=self._trigger, payload=payload)
+
+    def _download_and_store_message_attachments(self, message):
+        """
+        Method which downloads the provided message attachments and stores them in a datasatore.
+
+        :rtype: ``list`` of ``dict``
+        """
+        attachments = message.attachments
```
manasdk: So the attachment is downloaded regardless of size anyway - correct?

We discard the attachment if it is larger than configured but will be in memory AFAICT.

manasdk:
LGTM

packs/email/sensors/imap_sensor.py
```diff
@@ -35,13 +47,17 @@ def setup(self):
     def poll(self):
         self._logger.debug('[IMAPSensor]: entering poll')
 
-        for name, mailbox in self._mailboxes.items():
-            self._poll_for_unread_messages(name, mailbox)
+        for name, values in self._mailboxes.items():
```
manasdk: six?

packs/email/README.md
Kami: Clarified units in the readme.

packs/email/sensors/imap_sensor.py
Kami: Yeah, sadly that's how the underlying library works - the attachment is still downloaded, but we don't store it in the datastore.

Not ideal, but we can look into optimizing it (not downloading the attachment) later.

packs/email/sensors/imap_sensor.py
Kami: We don't consistently use `six` in st2contrib, we should probably consider adding `six` to the base requirements list which are installed with every pack.

packs/email/sensors/imap_sensor.py
Kami: Clarified that in the readme.

PR #226, opened 2015-06-10
WIP: New Slack actions

DoriftoShoes:
This PR adds new Slack actions to support automating user invites.
#### Actions:
- send_invite
- auth_test
- list_users
- list_channels

### DIFF
--- a/packs/slack/actions/auth_test.py
+++ b/packs/slack/actions/auth_test.py
@@ -1,5 +1,3 @@
-import json
-import httplib
 import urllib
 
 import requests
@@ -25,11 +23,11 @@ class AuthTestAction(Action):
 
         data = urllib.urlencode(body)
         response = requests.get(url=url,
-                                 headers=headers, params=data)
+                                headers=headers, params=data)
 
         results = response.json()
 
-        if results['ok'] == True:
+        if results['ok'] is True:
             return results
         else:
             failure_reason = ('Failed to authenticate: %s \
--- a/packs/slack/actions/auth_test.py
+++ b/packs/slack/actions/auth_test.py
@@ -27,8 +27,10 @@ class AuthTestAction(Action):
         response = requests.get(url=url,
                                  headers=headers, params=data)
 
-        if response.status_code == httplib.OK:
-            return response.json()
+        results = response.json()
+
+        if results['ok'] == True:
+            return results
         else:
             failure_reason = ('Failed to authenticate: %s \
                               (status code: %s)' % (response.text,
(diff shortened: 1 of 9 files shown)


packs/slack/actions/auth_test.py
```diff
@@ -0,0 +1,39 @@
...
+            'token': token
+        }
+
+        data = urllib.urlencode(body)
+        response = requests.get(url=url,
+                                headers=headers, params=data)
+
+        results = response.json()
+
+        if results['ok'] is True:
```
lakshmi-kannan: `is True` is not needed.

packs/slack/actions/auth_test.py
```diff
@@ -0,0 +1,39 @@
...
+            return results
+        else:
+            failure_reason = ('Failed to authenticate: %s \
+                              (status code: %s)' % (response.text,
+                              response.status_code))
+            self.logger.exception(failure_reason)
+            raise Exception(failure_reason)
```
lakshmi-kannan: heh dude. you should really use `return early` approach.

packs/slack/config.yaml
```diff
@@ -1,4 +1,26 @@
 ---
+# Used for auth test action
+auth_test_action:
+  token: ""
+
+# Used for list channels
+list_channels_action:
```
Kami: To me it kinda seems that most of those configuration options (except the token) should really just be action parameters.

Kami: Edit: I see user can also specify it as a parameter and those values there work more as a default value. I think I'm fine with that.

packs/slack/actions/list_channels.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: "list_channels"
+  runner_type: "run-python"
+  description: "List channels in a Slack Org"
+  enabled: true
+  entry_point: "list_channels.py"
+  parameters:
+    token:
+      type: "string"
+      description: "Slack API token."
+    exclude_archived:
```
Kami: This should be a boolean and we should send a correct integer inside the action.

packs/slack/actions/send_invite.yaml
```diff
@@ -0,0 +1,26 @@
...
+  parameters:
+    email:
+      type: "string"
+      description: "Email address to send invitation to."
+      required: true
+    first_name:
+      type: "string"
+      description: "Users first name"
+    channels:
+      type: "string"
```
Kami: This should be an array and we should build a comma delimited string inside an action.

Kami: It actually looks like you expect an array from a config, but not from a parameter :)

DoriftoShoes:
I'm closing this PR as I moved these changes over to a new repo with auto generated actions.

PR #227, opened 2015-06-10
Don't hardcode post_result chanel to "hubot"

armab:
Playing with ChatOps & Slack, Hubot pack.

The problem is that channel name where st2 results are sent is hardcoded to `hubot` [here](https://github.com/StackStorm/st2contrib/blob/master/packs/hubot/rules/notify_hubot.yaml#L11). It means if channel `hubot` wasn't created in Slack - you don't get any output.

As I understand `ST2_CHANNEL` env here: https://github.com/StackStorm/hubot-stackstorm#plugin-environment-variable-options is that special channel name where st2 command results should go.

With provided fix if you post in `#general` channel, st2 output should go into that defined #`ST2_CHANNEL`.

---
P.S. Not sure if this fix doesn't breaks any other use cases, maybe there is specific purpose?


manasdk:
@armab That is not how this works. The channel in the criteria is the `st2_channel` for notifications and for better or worse has nothing to do with a specific chat client.

Checkout https://github.com/StackStorm/st2/blob/a0e9b3d00a003dc41493d212763cd48fec8d62a7/instructables/chatops.md

> As I understand ST2_CHANNEL env here: https://github.com/StackStorm/hubot-stackstorm#plugin-environment-variable-options is that special channel name where st2 command results should go.

That is the stackstorm channel used for notification as defined by the rule. If you are trying to deploy that value should also be `hubot`.

See https://github.com/StackStorm/hubot-stackstorm/blob/master/scripts/stackstorm.js#L48

Btw, -1 to the PR.

manasdk:
The chat client channel is opaque and adapter specific. 

In case of slack(which is what we use for testing pretty heavily) the bot is already in already active in a specific channel and/or has access to a specific channel. When a user fires of a request the slack channel will be automatically inferred and response will be in the same channel. Therefore, the user often does not have to worry about channels for the most part.

Kami:
I think we should also put some of that clarification into `hubot-stackstorm` README and into StackStorm docs (although, IIRC, we don't have any public facing docs for notifications right now).

At first I was also a bit confused about the channel terminology (Slack channel vs StackStorm notification channel, maybe we should use more explicit terms?)

armab:
In [plugin environment variables](https://github.com/StackStorm/hubot-stackstorm#plugin-environment-variable-options) it's clearly said: 
> ST2_CHANNEL - Notification channel where all the notification messages should be sent to. If you use Slack adapter, that's the name of the Slack channel.

It means if I set that env and use Slack - all messages should be sent into that specific Slack channel. No?

I look on this from the user's side:

### Reproduce №1
1. Create Slack account, create `#chatops` channel
2. Enable Hubot integration in Slack, generate token
3. Run Hubot with st2 plugin, evn `ST2_CHANNEL=chatops`
4. In Slack invite hubot into `#chatops` channel
5. Run in Slack channel `#chatops` command: `!pack info ansible `

### Reproduce №2
1. Create Slack account (don't create any new channels and make sure #hubot channel doesn't exist)
2. Enable Hubot integration in Slack, generate token
3. Run Hubot with st2 plugin, evn `ST2_CHANNEL=hubot`
4. In Slack invite hubot into `#general` channel
5. Run in Slack channel `#general` command: `!pack info ansible`

### What I get (no output): 


### What is expected (output): 


### Questions

* How should I guess as user that `#hubot` channel should be created to get any output?
* What should I do if I want to recieve all messages from st2 in specific channel (as `ST2_CHANNEL` doc said)?

--------

In `packs/hubot/rules/notify_hubot.yaml` :
* [`trigger.channel`](https://github.com/StackStorm/st2contrib/pull/227/files#diff-5762ae07bbeca0832c5512badde92beeL10) refers to `ST2_CHANNEL`
* [`trigger.data.source_channel`](https://github.com/StackStorm/st2contrib/pull/227/files#diff-5762ae07bbeca0832c5512badde92beeL16) refers to current channel where user posted his chatops command

manasdk:
> In plugin environment variables it's clearly said:

>> ST2_CHANNEL - Notification channel where all the notification messages should be sent to. If you use Slack adapter, that's the name of the Slack channel.

> It means if I set that env and use Slack - all messages should be sent into that specific Slack channel. No?

That is wrong. I will fix the description. Again `ST2_CHANNEL` has nothing to do with the slack client. We should have called it something else - like `ST2_NOTIFICATION_TARGET`. It can take values like `hubot`, `email` etc. and defines the notification channel that StackStorm would like the notify. The notification channel is defined by a rule.

manasdk:
@armab 

Looking at `Reproduce №2` - did you install the `hubot` pack from https://github.com/StackStorm/st2contrib/tree/master/packs/hubot? Specifically the rule `https://github.com/StackStorm/st2contrib/blob/master/packs/hubot/rules/notify_hubot.yaml` should be enabled.

If so take a look at the execution list and check if there is any failure posting back to hubot. The action that should execute is `https://github.com/StackStorm/st2contrib/blob/master/packs/hubot/actions/post_result.yaml`

armab:
Ok, that `ST2_CHANNEL` description is what confused me totally.

Then there is still bug with `hubot-stackstorm` (for Reproduce №2 I've got error in Hubot logs, unless I create `#hubot` channel in Slack), I'll report in that repo.

manasdk:
@armab See https://github.com/StackStorm/hubot-stackstorm/pull/8

manasdk:
> Then there is still bug with hubot-stackstorm (for Reproduce №2 I've got an error in Hubot logs and no output from st2, unless I create #hubot channel in Slack), I'll report in that repo.

Do report the issue - I would like to see the error and figure out the issue.

armab:
The error I got in Hubot logs was: `ERROR Unable to decode JSON: TypeError: Cannot call method 'send' of undefined`.

But the real reason of error was that I several times edited manually the rule `rules/notify_hubot.yaml` and it wasn't re-parsed (my wrong here). Things I did to re-parse it: 
```sh
st2ctl restart
st2ctl reload
```
didn't helped (a bit confusing too). 
Only `st2ctl reload --register-rules` worked. So I got wrong results.

Thanks @manasdk for your help :+1:

Kami:
Re description in `hubot-stackstorm` - yeah, that was my mistake.

Kami:
>  We should have called it something else - like ST2_NOTIFICATION_TARGET. It can take values like hubot, email etc. and defines the notification channel that StackStorm would like the notify. The notification channel is defined by a rule.

Yeah, that would make it a bit less confusion.

PR #229, opened 2015-06-15
Create empty config.yaml for ansible pack

armab:
When pack already exists and installed,
reinstall `st2 run packs.install packs=ansible` fails with error:
```
IOError: [Errno 2] No such file or directory: u'/opt/stackstorm/packs/ansible/config.yaml'
```

This PR should fix it.


lakshmi-kannan:
+1

PR #230, opened 2015-06-18
Better message formatting for Hubot

armab:
This enhances `post_result` message formatting for Hubot.

Replaces `yaml` with [`pyaml`](https://github.com/mk-fg/pretty-yaml) library which chooses more human-readable ways of formatting depending on situation.

#### Before: 
chatops holy cow broken

#### After:
chatops holy cow!

### DIFF
--- a/packs/hubot/actions/post_result.py
+++ b/packs/hubot/actions/post_result.py
@@ -2,7 +2,7 @@ import json
 import httplib
 import requests
 import six
-import yaml
+import pyaml
 from six.moves.urllib.parse import urljoin
 
 from st2actions.runners.pythonrunner import Action
@@ -13,7 +13,7 @@ __all__ = [
 
 
 def _serialize(data):
-    return yaml.safe_dump(data, default_flow_style=False)
+    return pyaml.dump(data)
 
 
 def format_possible_failure_result(result):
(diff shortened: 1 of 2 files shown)


packs/hubot/actions/post_result.py
```diff
@@ -2,7 +2,7 @@
 import httplib
 import requests
 import six
-import yaml
```
lakshmi-kannan: I didn't think there was a yaml library shipped with python by default. This shouldn't have worked with pyyaml. It appears there is a pyyaml and a pyaml. pyaml looks like the one you want.

packs/hubot/requirements.txt
```diff
@@ -1,2 +1,3 @@
 requests>=2.5.1,<2.6
 six
```
lakshmi-kannan: I wonder how things worked without a pyyaml dep before.

packs/hubot/requirements.txt
```diff
@@ -1,2 +1,3 @@
...
+pyaml
```
lakshmi-kannan: Nit: Alpha order?

packs/hubot/requirements.txt
armab: Yeah, good catch.

Seems it's just global requirement: https://github.com/StackStorm/st2/blob/a41fc5d95134b98947c456ff6d2a49435fc6f035/requirements.txt#L16

lakshmi-kannan:
LGTM

Kami:
Nice :+1:

PR #233, opened 2015-06-28
Bitbucket pack

Aamir-raza-1:


packs/bit-bucket/config.yaml
```diff
@@ -0,0 +1,4 @@
+Host: https://bitbucket.org/api/1.0/
```
Kami: Again, same as in the other PR - it's unlikely that this URL will ever change so you can just use a constant in the module - no need for it to be a configuration ion.

packs/bit-bucket/config.yaml
```diff
@@ -0,0 +1,4 @@
...
+username: <bitcuket-username>
+password: <your-bitbucket-password>
```
Kami: Looks like BitBucket does indeed support oAuth which is safer and better than using email, but I'm fine with using password for now since oAuth token flow is a big complicated and requires user interaction.

packs/bit-bucket/requirements.txt
```diff
@@ -0,0 +1 @@
+requests>=2.6.0
```
Kami: Did you consider using https://pypi.python.org/pypi/bitbucket-api?

Kami:
`pack.yaml` file needs to contain metadata and can't be empty.

packs/bit-bucket/requirements.txt
Aamir-raza-1: @Kami, Yep I looked at it, but then went for using REST API, because I though the library might be limited in functionality and if we further need to extend the pack then we can extend it just by adding a call to API endpoint rather than waiting for the python package to be upgraded to the latest release for incorporating the latest changes.

jfryman: What was missing in the upstream module that is not in the API? Looks like several good actions in the official API that aren't quite in here yet.

Kami: I agree with @jfryman.

In addition to that, upstream library already implements oauth support - Implementing that ourselves would be time consuming and a waste of time so we should just use the upstream library.

Aamir-raza-1: Ok, I will use that and then generate a new PR. Going for the bit-bucket API library.

Kami: Sounds good. You can also re-use this branch if that makes it easier.

Aamir-raza-1:
I am working on more actions, Once they are done, I will push them upstream.

packs/bit-bucket/pack.yaml
```diff
@@ -0,0 +1,7 @@
+name: BitBucket
+description: Pack which allows integration with BitBucket.
+keywords:
+ - bitcuket
```
Kami: Typo, `bitcuket` -> `bitbucket` :)

packs/bit-bucket/requirements.txt
```diff
@@ -0,0 +1 @@
+bitbucket-api-0.5.0
```
Kami: This is not a valid Python version specific. Should be something like that: `bitbucket-api>=0.5.0,<0.6`

packs/bit-bucket/actions/lib/action.py
```diff
@@ -0,0 +1,17 @@
+from st2actions.runners.pythonrunner import Action
+from bitbucket.bitbucket import Bitbucket
+
+
+class BitBucketAction(Action):
+    def __init__(self, config):
+        super(BitBucketAction, self).__init__(config)
+
+    def perform_request(self, repo=''):
```
Kami: This method name is slightly confusing since it doesn't perform an actual http request - it just instantiates a client.

Calling it `_get_client` or similar would be less confusing, imo.

packs/bit-bucket/actions/archive_repo.yaml
```diff
@@ -0,0 +1,10 @@
+name: archive_repo
+runner_type: run-python
+description: Download a repository as an archive
+enabled: true
+entry_point: archive_repo.py
+parameters:
+  repo:
+    type: string
+    description: provide repositroy name for downloading as archive
```
Kami: Typo in `repositroy`

packs/bit-bucket/actions/archive_repo.yaml
```diff
@@ -0,0 +1,10 @@
+name: archive_repo
+runner_type: run-python
+description: Download a repository as an archive
```
Kami: We don't have a schema for the result object right now so it would be good to indicate that this action returns a path to the downloaded archive on success.

packs/bit-bucket/actions/create_issue.py
```diff
@@ -0,0 +1,17 @@
+from lib.action import BitBucketAction
+
+
+class CreateIssueAction(BitBucketAction):
+    def run(self, **kwargs):
```
Kami: Please explicitly declare the arguments here `def run(self, repo, title, ...)`

packs/bit-bucket/actions/create_issue.yaml
```diff
@@ -0,0 +1,26 @@
...
+    required: true
+  desc:
+    type: string
+    description: Description of the issue to create
+    required: true
+  status:
+    type: string
+    description: status of the issue, for example, bug, enahancement, proposal
+    required: true
+  kind:
```
Kami: Looks like this could be an enum. See https://github.com/StackStorm/st2contrib/blob/master/packs/smartthings/actions/set_temperature.yaml#L8 for an example.

packs/bit-bucket/actions/create_repo.yaml
```diff
@@ -0,0 +1,10 @@
+name: create_repo
+runner_type: run-python
+description: Create a Repository
+enabled: true
+entry_point: create_repo.py
+parameters:
+  repo:
+    type: string
+    description: provide repositroy name you want to create
```
Kami: Same typo here.

packs/bit-bucket/actions/create_service.yaml
```diff
@@ -0,0 +1,18 @@
+name: create_service
+runner_type: run-python
+description: Create a service/hook for a given repository
+enabled: true
+entry_point: create_service.py
+parameters:
+  repo:
+    type: string
+    description: provide repositroy name for listing the branches
```
Kami: Description is not accurate for this action.

packs/bit-bucket/actions/create_service.yaml
```diff
@@ -0,0 +1,18 @@
...
+    required: true
+  url:
+    type: string
+    description: provide repositroy name for listing the branches
```
Kami: Same here.

packs/bit-bucket/actions/create_ssh_key.yaml
```diff
@@ -0,0 +1,18 @@
+name: create_ssh_key
+runner_type: run-python
+description: Create SSH key
```
Kami: I needed to check the Python library documentation to figure out what this action does.

Imo, we should change the description so it says something like "Associate an ssh key with your account and return it." since this method / action associated an existing public key with your account.

packs/bit-bucket/actions/create_ssh_key.yaml
```diff
@@ -0,0 +1,18 @@
...
+enabled: true
+entry_point: create_ssh_key.py
+parameters:
+  repo:
+    type: string
+    description: Requires repo name
+    required: true
+  ssh_key:
+    type: string
+    description: Generated ssh-key
```
Kami: Would be better to say something like "Public SSH key to associate with your account".

packs/bit-bucket/actions/delete_issue.yaml
```diff
@@ -0,0 +1,14 @@
...
+enabled: true
+entry_point: delete_issue.py
+parameters:
+  repo:
+    type: string
+    description: provide repositroy name for listing the branches
+    required: true
+  id:
+    type: array
+    description: ID of the issue to delete
```
Kami: Confused about this one. 

It looks like you pass in an array and not a single id. In this case, the action should be called `delete_issues` and the argument should be called `ids`. As an alternative, you can change the type to `string` and update the action so it operates with a single id and not a list.

packs/bit-bucket/actions/delete_service.yaml
```diff
@@ -0,0 +1,14 @@
...
+enabled: true
+entry_point: delete_service.py
+parameters:
+  repo:
+    type: string
+    description: provide repositroy name
+    required: true
+  id:
+    type: array
+    description: ID of the service to delete
```
Kami: Same comes (delete_issue) applies here.

packs/bit-bucket/actions/list_branches.py
```diff
@@ -0,0 +1,15 @@
...
+class ListBrachesAction(BitBucketAction):
+    def run(self, repo):
+        """
+        List Braches of Repository with author names and message
+        """
+        bb = self.perform_request(repo=repo)
+        success, result = bb.get_branches()
+        branches = {}
+        for key, value in result.iteritems():
+            branches[key] = {'author': value['author'],
```
Kami: Does the result object include more information about each branch?

If so, we should also include other available info.

packs/bit-bucket/actions/list_issues.py
```diff
@@ -0,0 +1,19 @@
...
+    def run(self, repo):
+        """
+        List Issues of Repository with title
+        of the issue its status and reporter
+        """
+        bb = self.perform_request(repo=repo)
+        success, result = bb.issue.all()
+        issues = {}
+        for i in range(result['count']):
+            issues[result['issues'][i]['title']] = {
```
Kami: Does the title need to be unique?

If it doesn't we should return a list instead of an object (dict).

Kami: Also, similar as above, we could simple return the whole issue object as returned by the API instead of only picking a couple of keys / attributes from it.

packs/bit-bucket/actions/list_ssh_keys.yaml
```diff
@@ -0,0 +1,10 @@
+name: list_ssh_keys
+runner_type: run-python
+description: List all SSH keys for a user
+enabled: true
+entry_point: list_ssh_keys.py
+parameters:
+  repo:
```
Kami: Hm, this action performs operation on an user account and not on a repository. Does it still need the repo argument?

packs/bit-bucket/actions/list_user_repo.py
```diff
@@ -0,0 +1,21 @@
+from lib.action import BitBucketAction
+
+
+class GetRepos(BitBucketAction):
```
Kami: This action returns a list so please call it `ListRepos` and `list_repos` (same as you do with `list_branches`).

packs/bit-bucket/actions/list_user_repo.py
```diff
@@ -0,0 +1,21 @@
...
+        password in config file. It returns repo's
+        name, its state and private status(returns
+        True if its private, False otherwise)
+        """
+        repository = []
+        bb = self.perform_request()
+        success, repos = bb.repository.all()
+        for repo in sorted(repos):
+            repository.append(
+                {repo['name']: {'state': repo['state'],
```
Kami: You should decide to either return a list of repo objects or an object (dict) where the key is a repo name and the value are other repo attributes.

It makes little sense to have another level of indirection and return a list of nested objects (nesting under repo name in this case doesn't provide any value).

packs/bit-bucket/actions/list_user_repo.yaml
```diff
@@ -0,0 +1,5 @@
+name: list_user_repo
```
Kami: list_user_repo -> list_user_repos.

packs/bit-bucket/actions/list_issues.py
Aamir-raza-1: @Kami, Yeah, we can return the whole object, it contains enormous details. I only added the ones that the user would need to further interact with the other actions. Returning the object would be easier for me though.

packs/bit-bucket/actions/create_issue.yaml
Aamir-raza-1: Yep, I am changing it to enum

packs/bit-bucket/actions/list_ssh_keys.yaml
Aamir-raza-1: No, I will update it. Thanks

Kami:
@Aamir-raza-1 How is this looking? Did you address all the issues and can I review it again?

Thanks

armab:
Also maybe `bitbucket` is more expected naming for pack instead of `bit-bucket`?

Kami:
Unlike with Travis CI, Bitbucket is a single world so we should just name the pack (and directory) `bitbucket` :)

Aamir-raza-1:
Yeah sure, @Kami You can review the chages, I have fixed them.

Kami:
I've fixed and improved some descriptions and merged pack into master.

The original pull request didn't include a readme so please create a new pull request based off master which adds a readme.

Thanks.

Aamir-raza-1:
ok sure, I will add a readme. Thanks

Kami:
On a related note - I tested the actions, but I couldn't get "list_issues" and some other to work.

The config contains correct values (username, password and email), but `list_issues` and some other return `Service not found.` error back.

Looking at the bitbucket-api Python library, this error is returned if API returns status code which is >= 400 < 500.

Any ideas what could be causing this and how to get it to work?

Aamir-raza-1:
When there are no issues to list and you are using `list_issues`, it responds with the message `service not found`, similar is the case with some other actions that you are referring to.

Kami:
Hm, that's rather weird error to return in that case. We should consider submitting a fix / improvement to bitbucket-api project.

In any case, that's for the clarification. I will try with a repo which has some issues and see what happens.

Aamir-raza-1:
I will improve that to return proper message in case there are no issues, and replicate it to relevant actions

PR #234, opened 2015-07-04
Travis CI pack

Kami:
Fixes on top of #218


Kami:
I fixed all the bugs and addressed all the issues. I will merge this into master, but we still need (better) documentation on how to obtain the access token.

Aamir-raza-1:
Let me see, the way of obtaining token, I'lll update today here

Kami:
@Aamir-raza-1 Alright, please open a new PR for those changes.

Aamir-raza-1:
@Kami We need to make changes to README. We first need to generate a Github token. Once that is generated we will send a `POST` request to `/auth/github` API and it will return us a token of Travis CI API. That token would be used for future corresponding requests. So we need github token just for getting the Travis CI token. 
I already have mentioned the way to generate github API token in README with images, We just need to add the code to get Travis CI token (a POST call)
What do you suggest?

Kami:
@Aamir-raza-1 Yeah, in this case the user would only need to specify a Github token in the config file and we would dynamically retrieve Travis CI access token using it.

One thing I'm not sure is (need to check the docs), if the /auth/github flow is possible without user-interaction (e.g. requiring user to open a browser, etc.). If it's not possible, then we *can't* do this automatically and we will need to write the instructions on how to do that and put it in the readme.

If it's possible in this case, we can just go with that approach then - simply retrieve a new Travis CI access token on every action invocation - it's not ideally (ideally, we would re-use / cache the token), but it works fine for now.

Aamir-raza-1:
@Kami , Let me see, I guess the flow looks like this to me.
we need to use git API to first grant access to the travis CI to user account, after access is granted we will retrieve client id and client secret for getting git api key and then use that api key to get travis api key. What do you suggest.

Kami:
Looks like Github auth token should work:

> If you have a GitHub token, you can use the GitHub authentication endpoint to exchange it for an access token. The Travis API server will not store the token or use it for anything else then verifying the user account.

> Note that therefore the API cannot be used to create new user accounts. The user will have to log in at least once via the web interface before interfacing with the API by other means.

Aamir-raza-1:
Hmm, Let me look what can we do with this.

Aamir-raza-1:
@Kami, How can I get the latest code that you have made changes to. When i clone it clones it from my fork and I have the same old code. I need the one that has been changed so that I can start making changes to. Thanks

Kami:
@Aamir-raza-1 Go to master, pull down the latest changes and include your changes in a new branch based of master:

```bash
git checkout master
git pull origin master
git checkout -b my_new_branch_with_auth_changes
....
```

Aamir-raza-1:
@Kami, Thats what I have already tried. Still the code changes that you have made are not visible and the actions travisci action also not cloned. 
when i pull it says your branch is upto date with master.
Let me see it. Thanks

Kami:
@Aamir-raza-1 you are probably pulling from your fork repo and not upstream repo. Do:

```bash
git remote add upstream https://github.com/StackStorm/st2contrib.git
git pull upstream master
...
```

Kami:
@Aamir-raza-1 Figured it out?

Aamir-raza-1:
Will work on it in sometime, I will sync my fork and add remote. Thanks

Aamir-raza-1:
I have synced my fork got the latest code, I am now going to write a function to retrieve travis API token, we need to have GIT api token in config file.

Kami:
@Aamir-raza-1 Yeah that's fine. Previously I removed it since it wasn't used in the code.

Aamir-raza-1:
@Kami, Hey sorry for the delay, I was off for EID holidays. I have created a PR. Please see https://github.com/StackStorm/st2contrib/pull/252
steps are that you need to supply a username and password for your git. 
I wrote a action that will use that username and password and retrieve a token from git. then that token would be used fo get token from travis. so the process is autmated a bit now.

PR #235, opened 2015-07-07
Update Twitter Search Sensor

jfryman:
This PR updates the twitter sensor to be able to accept multiple complex search terms. `query` becomes an array, and the user is able to have more than one query going during a poll.

### DIFF
--- a/packs/twitter/sensors/twitter_search_sensor.py
+++ b/packs/twitter/sensors/twitter_search_sensor.py
@@ -28,7 +28,8 @@ class TwitterSearchSensor(PollingSensor):
         self._last_id = None
 
         if type(self._config['query']) is not list:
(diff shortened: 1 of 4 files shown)


manasdk:
+1

Kami:
LGTM.

packs/twitter/sensors/twitter_search_sensor.py
```diff
@@ -29,7 +29,7 @@ def setup(self):
 
     def poll(self):
         tso = TwitterSearchOrder()
-        tso.set_keywords([self._config['query']])
+        tso.set_keywords(self._config['query'])
```
Kami: Since we don't have a schema for the config values, we should throw inside the constructor if the `query` value is not an array to prevent confusion for the users who are upgrading the pack.

PR #236, opened 2015-07-10
Adds Publish action to RabbitMQ

jtopjian:
This commit adds a publish action to the RabbitMQ pack, allowing a crafted message to be published to a RabbitMQ server.

### DIFF
--- a/packs/rabbitmq/actions/publish_message.py
+++ b/packs/rabbitmq/actions/publish_message.py
@@ -1,0 +1,15 @@
+import pika
+
+from st2actions.runners.pythonrunner import Action
+
+
+class PublishMessageAction(Action):
+    def run(self, host, port, username, password, virtual_host, exchange, exchange_type,
+            exchange_durable, routing_key, message):
+        credentials = pika.PlainCredentials(username, password)
+        parameters = pika.ConnectionParameters(host, port, virtual_host, credentials)
+        connection = pika.BlockingConnection(parameters)
+        channel = connection.channel()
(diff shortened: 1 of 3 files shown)


packs/rabbitmq/actions/publish_message.py
```diff
@@ -0,0 +1,13 @@
+from kombu import BrokerConnection, Exchange, Queue
```
manasdk: This pack is built around pika. Using kombu probably works as it is a core stackstorm dependency but it is perhaps better to use pika.

Something like https://www.rabbitmq.com/tutorials/tutorial-one-python.html

Pika generally makes for simpler APIs than kombu. In core stackstorm also we are considering a move towards pika.

jtopjian: Yeah, I chose Kombu since it's required by StackStorm. I'll take a look at switching over to Pika.

jtopjian:
@manasdk switched to Pika. Let me know how this looks. I can squash the commits to something cleaner before the final merge.

manasdk:
Looks good. Thanks for switching to Pika.

Travis is complaiing about line length -
/home/travis/build/StackStorm/st2contrib/packs/rabbitmq/actions/publish_message.py:6:101: E501 line too long (129 > 100 characters)

Will merge after that gets fixed.

jtopjian:
@manasdk looks like everything is checking out. Let me know if there are any other changes to make.

manasdk:
Done. Merged!

Kami:
Nice :+1:

PR #239, opened 2015-07-14
post instead of get for login, add a missing config for list_events

Itxaka:
Fix for the mmonit pack, now that I have a stackstorm and mmonit instance running I can find any issues with them :)


jfryman:
Stellar. Thanks for the code!

PR #241, opened 2015-07-14
wrong config name parameter

Itxaka:
Sorry, dumb error on my part, didnt even noticed!

Can confirm that sensor is working now.


manasdk:
LGTM.

PR #242, opened 2015-07-15
Mmonit sensor refactor and bugfixing

Itxaka:
- add 2 types of filter for a more granulated control of events to listen
 - refactor sensor so it uses the datastore inteligently and does not spam the same events, in case of long standing events
 - clear the datastore when the events dissapear
 - add a logger for debugging
 - log why the events are not being triggered so nobody gets crazy on why they are missing events!
 - make quotes be consisteng across the whole file
 - fix the login using a get instead of the proper post


jfryman:
:metal: thanks for the code!

PR #244, opened 2015-07-15
IMAP Sensor fixes

jfryman:
This PR fixes two issues with the IMAP sensor.

1) IMAP sensor dies when attempting to read an older message marked as "Unread"
2) IMAP sensor does not properly parse message headers

### DIFF
--- a/packs/email/config.yaml
+++ b/packs/email/config.yaml
@@ -1,16 +1,11 @@
-### Place configuration data for your pack here.
 ---
 imap_mailboxes:
-  default:
-    server: ""
+  steve_main:
+    server: imap.kolabnow.com
     port: 993
-    username: ""
-    password: ""
-    ssl: True
-    download_attachments: True
-
+    username: steve@fryman.io
+    password: bsKpcHdoYQTAmnozhABJWipLxDdmYULXg9stDhti4jPKZdvLih
+    ssl: true
 max_attachment_size: 1024
 attachment_datastore_ttl: 1800
 
-smtp_listen_ip: '127.0.0.1'
-smtp_listen_port: 25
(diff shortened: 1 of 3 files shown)


packs/email/config.yaml
```diff
@@ -1,16 +1,11 @@
...
-    server: ""
+  steve_main:
+    server: imap.kolabnow.com
     port: 993
-    username: ""
-    password: ""
-    ssl: True
-    download_attachments: True
-
+    username: steve@fryman.io
```
Kami: I assume you didn't want to include this? :)

packs/email/sensors/imap_sensor.py
```diff
@@ -41,17 +41,18 @@ def __init__(self, sensor_service, config=None, poll_interval=30):
...
         self._logger.debug('[IMAPSensor]: entering setup')
 
-        if 'imap_mailboxes' in self._config:
-            self._parse_mailboxes(self._config['imap_mailboxes'])
-
     def poll(self):
         self._logger.debug('[IMAPSensor]: entering poll')
 
+        if 'imap_mailboxes' in self._config:
+            self._parse_mailboxes(self._config['imap_mailboxes'])
```
Kami: Just want to confirm - we re-establish a connection to the IMAP server on every poll now, right?

jfryman: Yes.

jfryman: Ok! After some :tea:, I am human again.

I put some details on this on commit https://github.com/StackStorm/st2contrib/commit/a5d6e76e6e60616ed5d51b46895cbe89d5bf5a4a. I was having to do deep traces into the Python IMAP library and how it implemented RFCs as I was getting protocol errors. Time to fix vs overhead, I went this way to save some time. If we need to optimize for a long poll, we'll have to ditch the `easyimap` library and get closer to the library and protocol.

PR #245, opened 2015-07-16
FireEye CM/AM Appliance Integration

jfryman:
# FireEye Integration Pack

StackStorm integration with FireEye CM and AX threat intelligence devices

## Configuration

* `api_url` - HTTPS endpoint of FireEye CM appliance. (e.x.: https://fqdn.to.device)
* `username` - FireEye username
* `password` - FireEye password

## Actions

* `fireeye.get_alert_query`         - Request existing alert profiles with optional filters
* `fireeye.get_submission_results`  - Query results of completed job
* `fireeye.get_submission_status`   - Query status of running job
* `fireeye.submit_malware`          - Submit a Malware object to FireEye AX appliance
* `fireeye.view_ax_config`          - Returns a list of profiles and applications on AX devices


packs/fireeye/actions/submit_malware.py
```diff
@@ -0,0 +1,16 @@
+from lib.actions import BaseAction
+import json
+import os
+
+
+class SubmitMalwareAction(BaseAction):
+    def run(self, **kwargs):
```
Kami: Looking at the action metadata file, it also defines `file_path` and `file_type` parameters.

I assume we don't want to send those in the payload? Or it doesn't matter and it still works even if those parameters are included?

packs/fireeye/actions/submit_malware.py
```diff
@@ -0,0 +1,16 @@
...
+        payload = json.dumps(kwargs)
+        file_basename = os.path.basename(kwargs['file_name'])
```
Kami: Looks like metadata only defines `file_path` and not `file_name` parameter.

packs/fireeye/actions/submit_malware.py
jfryman: Yes sir. Those do need to be in the payload. They are set as mandatory in the action metadata.

packs/fireeye/actions/submit_malware.py
jfryman: Good :eyes:. Updated the metadata.

PR #247, opened 2015-07-16
Get rid of pyaml to pretty dump json.

manasdk:
* Using the yaml dumper only leads to more escaping and further formatting misery
  so user simpler technique and let original formatting take its course.

This fixes issues https://github.com/StackStorm/hubot-stackstorm/issues/24

screen shot 2015-07-16 at 4 00 15 pm


jfryman:
Crazy.

/me makes note to never `pyaml`.

Kami:
:+1:

PR #248, opened 2015-07-20
Fix a bug with special characters in slack.post_message action

Kami:
This pull request fixes a bug in the slack.post_message action - we not correctly escape the request body so it also works if the message contains special characters.

### DIFF
--- a/packs/slack/actions/post_message.py
+++ b/packs/slack/actions/post_message.py
@@ -1,6 +1,11 @@
 import json
 import httplib
 
+try:
+    from six.moves.urllib.parse import urlencode
+except ImportError:
+    from urllib import urlencode
+
 import requests
 
 from st2actions.runners.pythonrunner import Action
@@ -32,7 +37,8 @@ class PostMessageAction(Action):
         if disable_formatting:
             body['parse'] = 'none'
 
-        data = 'payload=%s' % (json.dumps(body))
+        data = {'payload': json.dumps(body)}
+        data = urlencode(data)
         response = requests.post(url=config['webhook_url'],
                                  headers=headers, data=data)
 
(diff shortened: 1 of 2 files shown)


packs/slack/actions/post_message.py
```diff
@@ -1,6 +1,11 @@
 import json
 import httplib
 
+try:
+    from six.moves.urllib.parse import urlencode
```
Kami: `six` is only available in the sandbox in v0.12dev.

(Well, to be accurate, it's should also be available now because we use "--system-site-packages" workaround which we should eventually get rid of)

manasdk:
I am cool with this change but we should also look at depreciating the slack pack in st2contrib. St2-slack has any more features 

Kami:
@manasdk I actually had no idea st2-slack exists, lol.

But yeah, as soon as the actions in st2-slack are "stable" we should merge them into st2contrib. For backward compatibility reasons we should leave this action here for a while though (or just add an action metadata definition for an existing action which calls the new run.py script - this way we don't need to maintain two Python scripts).

manasdk:
I don't think plan is to merge to st2contrib. This is where the slack pack will live in a separate repo. It is a way to switch to 1 pack per repo and eventually a pack forge

DoriftoShoes:
I definitely think we need to push down the route of one pack per repo.  So much cleaner.

Kami:
Yeah, I'm definitely for one pack per repo approach (it makes a lot of things easier and better), but we have limited discovery capabilities right now when using that approach.

manasdk:
> but we have limited discovery capabilities right now when using that approach.

Yup. We need to make some headway into a forge or basic discoverability capabilities.

PR #249, opened 2015-07-22
Authentication support for st2 pack.

manasdk:
Support for auth tokens and some other cleanup

* auth token can be provided in config.yaml as `auth_token` property. If not provided will be picked up from action environment variable
* `base_url` for stackstorm is picked up from config. If not provided `ST2_ACTION_API_URL` provided in action environment variable will be used. If the `st2` pack is meant to be used to work with the current stackstorm install it is better to leave base_url empty.
* Some other cleanup to un-handle exceptions in creation of client objects. Better to throw so the action fails fast.


DoriftoShoes:
+1

jfryman:
:+1:

PR #252, opened 2015-07-28
git auth changes for token

Aamir-raza-1:


Aamir-raza-1:
@Kami Have you reviewed this? What are your suggestions on this one

Kami:
@armab I haven't had the time to do it yet.

In general I'm not a fan of storing password in the config. Is it possible to use a scoped Github auth token instead of username + password?

Aamir-raza-1:
Yeah you are right, but at first we wanted to overcome the manual procedure of adding the token, so that's done by getting a git token by using user credentials, and then using that git token to get a travis api key and then adding that key to the API requests. 
I will look into the scoped token thing that you said, that's valid point though.

Kami:
@Aamir-raza-1 Yeah, I was wondering the same thing - will using a scoped token require the user to perform a manual step (e.g. run a script and open a URL) to retrieve it?

If so, than it's less than ideal.

Another problem with automatically obtaining a token using username + password is that we would obtain a new token on each action invocation which is not ideal (and secure) since we would leave old tokens laying around.

Aamir-raza-1:
@Kami, I have looked around for a bit, but there is a manual step involved in the first step that you mentioned. 
Coming to the username and password, I have tried to get authorizations but that too requires basic authentication and that wont be possible without supplying username and password.
And i dont think we have to again and again generate token, we will generate once and that will be written in the configs file and then we would reference to that token for future correpsondence. When i get the token i store that in config file and then use that to get travis api key.

Kami:
@Aamir-raza-1 Alright, so lets go with the username and password then.

Can you please update README and config.txt example?

packs/travis_ci/actions/get_git_token.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class GetGitToken(TravisCI):
+    def run(self):
+        username = self.config['username']
+        password = self.config['password']
+        url = 'https://api.github.com/authorizations'
+        host = 'api.github.com'
+        ContentType = 'application/json'
+        data = '{"scopes": ["read:org", "user:email",\
```
Kami: Please use a dict and `json.dumps` to create this JSON payload. Creating JSON using a string is not robust and also unsafe in many scenarios.

packs/travis_ci/actions/get_git_token.py
```diff
@@ -0,0 +1,22 @@
...
+                "repo_deployment","repo:status", "write:repo_hook"],\
+                "note": "toke for travis api"}'
+        headers = {}
+        headers['Host'] = host
+        headers['Content-Type'] = ContentType
+        response = requests.post(url, data=data, auth=(username, password),
+                                 headers=headers)
+        res = yaml.load(response.content)
```
Kami: There is no need to use yaml, API returns JSON which is already parsed by requests and available in `response.json()`.

See https://github.com/StackStorm/st2contrib/commit/82b7df9589409c47e63f77a846f67e7f38519e25 for details.

Aamir-raza-1: duly noted sir.

packs/travis_ci/actions/get_git_token.py
Aamir-raza-1: Yes, I already did that, but did not push it. I will surely push that upstream.

Aamir-raza-1:
Yeah sure, I will update the README and config files

packs/travis_ci/README.md
```diff
@@ -3,6 +3,26 @@
 Pack for integration of Travis CI into StackStorm. The pack includes the
 functionality to perform actions on Travis CI through StackStorm.
 
+##Steps for using the pack
+
+### Authorizng Travis app
+Assuming that you have git account, now you need to authorize Travis
+application and grant access to you git authorized applcation.
```
lakshmi-kannan: authorized application. typo.

packs/travis_ci/actions/get_git_token.py
```diff
@@ -0,0 +1,23 @@
...
+                "repo_deployment", "repo:status", "write:repo_hook"],
+                "note": "token for travis api"}
+        headers = {}
+        headers['Host'] = host
+        headers['Content-Type'] = ContentType
+        response = requests.post(url, data=json.dumps(data),
+                                 auth=(username, password),
+                                 headers=headers)
+        res = response.json()
+        self.config['token'] = res['token']
```
lakshmi-kannan: I don't think this writes it to config file unless you have that code somewhere.

packs/travis_ci/actions/get_git_token.py
```diff
@@ -0,0 +1,23 @@
+from lib.action import TravisCI
+import json
+import requests
+
+
+class GetGitToken(TravisCI):
+    def run(self):
+        username = self.config['username']
+        password = self.config['password']
+        url = 'https://api.github.com/authorizations'
```
lakshmi-kannan: This could be in config. Also, I think github APIs are versioned? If so, we should use a specific version,

packs/travis_ci/actions/get_git_token.py
```diff
@@ -0,0 +1,23 @@
...
+        host = 'api.github.com'
+        ContentType = 'application/json'
+        data = {"scopes": ["read:org", "user:email",
+                "repo_deployment", "repo:status", "write:repo_hook"],
+                "note": "token for travis api"}
+        headers = {}
+        headers['Host'] = host
```
lakshmi-kannan: Is this needed?

packs/travis_ci/actions/get_git_token.py
lakshmi-kannan: saw @Kami's comment on the other PR. It's fine as a constant. Also, looks like they are unversioned. https://api.github.com/

packs/travis_ci/actions/get_git_token.py
Aamir-raza-1: Yes, we need to automate the git token procedure as much as we can. First, the user has to authorize the app to grant access, after that `we need the git token in order to get Travis CI token`.  Getting Git token was also manual and user interaction was needed. So, we wrote the action to automate that as well and the user interaction should be reduced to the minimal.

packs/travis_ci/actions/get_git_token.py
Aamir-raza-1: @lakshmi-kannan, Can you guide me where the code for reading config is written, so we can extend that to write it in configs, or do you have any better approach in mind

lakshmi-kannan: I don't think writing back to config is a good approach. We can leave the actions independent and chain them using action chain as a workflow so we can automate the whole pipeline. Does this make sense? 

http://docs.stackstorm.com/actionchain.html should get you started on that approach.

Aamir-raza-1: @Kami, Can you suggest what would be the best way around it. We have to get get travisci token from this action and store it in config so that it might be used for further requests that needs authentication.

dennybaa: Tokens and kinda similar stuff is appropriate to pass as ENVIRONMENT. This is a common approach. I dunno but stackstorm might support passing environment to an action or chain.

Kami: Yeah, I agree with @lakshmi-kannan.

I don't think writing it back into the config.txt is a good idea. One thing we could do is cache it in some temporary file on disk and if this file doesn't exist, we should simply retrieve a new token.

Aamir-raza-1: Instead of writing to a file, I guess we should store the token into some global variable and then access it all over in the action. 
Because it is needed only once, we could set it in global variable and then use it afterwards.

Kami:
Closing this since the changes are out of date now.

If the changes are still relevant, please open a new PR against latest master - thanks.

PR #253, opened 2015-07-29
Add retries to fetching of instances

DoriftoShoes:
There is a race condition with the instance creation where we sometimes request the instance status and get an error.  This is to make that a little more robust.

### DIFF
--- a/packs/aws/actions/lib/action.py
+++ b/packs/aws/actions/lib/action.py
@@ -49,6 +49,8 @@ class BaseAction(Action):
         state_list = {}
         obj = self.ec2_connect()
         eventlet.sleep(timeout)
+        instance_list = []
+
         for _ in range(retries + 1):
             try:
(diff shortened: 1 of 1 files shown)


packs/aws/actions/lib/action.py
```diff
@@ -45,16 +45,25 @@ def split_tags(self, tags):
...
-        for instance in obj.get_only_instances([instance_id, ]):
+        eventlet.sleep(timeout)
+        for _ in range(retries + 1):
+            try:
+                instance_list = obj.get_only_instances([instance_id, ])
+            except Exception:
+                self.logger.info("Waiting for instance to become available")
+                eventlet.sleep(timeout)
+
+        for instance in instance_list:
```
manasdk: instance_list could be undefined at this point if all attempts throw an exception. Assign it a default value "[ ]" outside the for loop

manasdk:
+1

PR #259, opened 2015-08-03
Rule to restart hubot every 24 hours

manasdk:
* Hubot teardown and bring up can be necessary so a rule to help with that every 24hours


jfryman:
👍

PR #266, opened 2015-08-11
New Relic sensor User-Agent

jfryman:
This PR adds a User-Agent to all New Relic actions in StackStorm.


Kami:
LGTM, +1

Although it would be better if the requests version is not hard-coded. Otherwise we will need to update it once a new version of requests is released :)

Kami:
It would also be worth considering providing some of that functionality / information to all the actions and sensors from the StackStorm side.

PR #267, opened 2015-08-13
Bump `gitpython` in Git Pack

jfryman:
During troubleshooting yesterday, we found that an older version of `gitpython` is causing an inability to clone repos from GitHub Enterprise. Bumping to latest version resolves this.


Kami:
LGTM :+1:

PR #269, opened 2015-08-16
New ipcam pack

Kami:
I have a couple of cheap "consumer grade" IP cams at home and I thought it would be fun to hook it up with StackStorm.

Currently, the pack only contains "capture_screenshot" action.


dzimine:
This will fit better on st2incubator.

manasdk:
I am still voting for a st2iot repo.

Kami:
Imo, until we improve the pack discovery, we should put all the packs which work into st2contrib - otherwise it's hard / impossible to find the pack.

manasdk:
Prove unambiguously by creating a much more obvious problem and then go solve it with better information.

I would infact say if we were to assume pack discovery is in place where would you like to place this pack?

Kami:
Eventually, when a proper API for pack management lands in master, we will probably want to have a repo per pack.

We are not there yet right now (and judging by our roadmap, priorities and workload we won't be there any time soon either) so it makes little sense to split it out now - as mentioned above, it makes pack discovery harder. README.md in this repo is the closest thing to the "registry" website we have right now.

And even when a proper "registry" website becomes available (I assume this will and should happen before the package management API) we should still keep everything in a single repo until we have a proper management API in place, because current `packs.` command don't cope that well with a real repo per pack model.

PR #272, opened 2015-08-18
Initial ServiceNow integration

jfryman:
# ServiceNow Integration Pack

This integration allows bi-directional communication between StackStorm and ServiceNow REST API

# Overview

This action provides the basic REST primitaves necessary to allow communication between StackStorm and ServiceNow. Ideally, this integration will be consumed in a site-specific integration pack that defines actions to specific ServiceNow specific business logic.

ServiceNow provides two videos to demonstrate how to setup Inbound and Outbound Web Service integrations.

* Inbound Integration - https://www.youtube.com/watch?v=EhxgEECd7mQ
* Outbound Integration - https://www.youtube.com/watch?v=WeeDW_iRM8k

An example pack has been included in this pack to show the integration of ServiceNow with these two integration examples.

# Setup
## Configuration

* `instance_name` - Upstream Instance Name (e.x.: stackstorm)
* `username` - Username of service account
* `password` - Password of service account

## Actions

* `servicenow.get` - Get an entry from a ServiceNow Table
* `servicenow.update` - Update an entry in a ServiceNow Table
* `servicenow.insert` - Insert an entry to a ServiceNow Table
* `servicenow.delete` - Delete an entry from a ServiceNow Table

### DIFF
--- a/packs/servicenow/actions/get.py
+++ b/packs/servicenow/actions/get.py
@@ -3,8 +3,6 @@ from lib.actions import BaseAction
 
 class GetAction(BaseAction):
     def run(self, table, query):
-        self.client.table = table
-        # pylint: disable=no-member
-
+        self.client.table = table  # pylint: disable=no-member
         response = self.client.get(query)
         return response
--- a/packs/servicenow/actions/get.py
+++ b/packs/servicenow/actions/get.py
@@ -4,5 +4,5 @@ from lib.actions import BaseAction
 class GetAction(BaseAction):
     def run(self, table, query):
         self.client.table = table  # pylint: disable=no-member
-        response = self.client.get(query)
+        response = self.client.get(query)  # pylint: disable=no-member
         return response
--- a/packs/servicenow/actions/get.py
+++ b/packs/servicenow/actions/get.py
@@ -3,9 +3,6 @@ from lib.actions import BaseAction
 
 class GetAction(BaseAction):
     def run(self, table, query):
-        try:
-            self.client.table = table
-            response = self.client.get(query)
-            return response
-        except e:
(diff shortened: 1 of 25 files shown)


packs/servicenow/actions/get.py
```diff
@@ -0,0 +1,11 @@
+from lib.actions import BaseAction
+
+
+class GetAction(BaseAction):
+    def run(self, table, query):
+        try:
+            self.client.table = table
+            response = self.client.get(query)
+            return response
+        except e:
```
Kami: Minor style thing - this `try / except` is not needed if we don't do anything special with the exception (e.g. log it or throw a more granular exception) since the default behavior without try / except block is the same.

Kami:
Looks like there are some lint issues which should be resolved when try / except block is removed - https://travis-ci.org/StackStorm/st2contrib/builds/76123390

packs/servicenow/actions/get.py
jfryman: Help me better understand how to return failure in the python runner. Is there a better way to do this other than returning generic exceptions?

Kami:
I would prefer to call the pack `service_now` sine the word is capitalized (minor style and consistency thing).

Besides that, LGTM, +1.

packs/servicenow/actions/get.py
Kami: We have no better way right now.

The pattern we use right now is throwing an exception with appropriate error message which causes action to fail.

I know this is not the most ideal approach since you can't that easily differentiate between different exception types (you need to check the message), but more or less the same problem would exist even if we used a different approach.

PR #278, opened 2015-08-25
Changing issue number in "github" pack to string to allow evaluation

emedvedev:
This was reported in st2web issues: https://github.com/StackStorm/st2web/issues/184
Changing validation type to string solves the problem, and I imagine that's a pretty common use case.

### DIFF
--- a/packs/github/actions/add_comment.yaml
+++ b/packs/github/actions/add_comment.yaml
@@ -14,7 +14,7 @@
       description: "Repository name."
       required: true
     issue:
-      type: "integer"
+      type: "string"
       description: "Issue / pull request number"
       required: true
     body:
--- a/packs/github/actions/get_issue.yaml
+++ b/packs/github/actions/get_issue.yaml
@@ -14,6 +14,6 @@ parameters:
     description: "Repository name."
     required: true
   issue_id:
-    type: integer
+    type: "string"
     description: Issue id
     required: true


emedvedev:
What bugs me about this is that we don't really have a validation model that's suitable for evaluation inside the action fields: we're checking for the field content, not the actual result of the evaluation. I think it's fine for now, but as we get wider adoption we'll probably going to have more issues like this one.

Kami:
Yeah, I agree that this is probably the best workaround for now (I don't see much value in the issue number being an int - you usually don't compare two issue numbers, etc). 

It's worth noting that in theory it's a backward incompatible change, but I think it's unlikely that someone would hard code an issue number in a workflow / action so it most likely won't affect anyone.

Kami:
On a related note - I believe there are some other actions in this pack which also take the issue number.

We should make sure the type is consistent in all the actions and that we perform a casting inside the action where necessary (IIRC, Github library we use requires us to pass in an int and not string in some places).

emedvedev:
There's a cast in `add_comment.py`. I'll check the rest of actions in a bit.

emedvedev:
Changed it in one more action for consistency and made sure there's casting in both actions.

emedvedev:
Travis failed because of a timeout, weird.

enykeev:
> weird

Not really. https://status.github.com/

screen shot 2015-08-25 at 3 40 13 pm

Kami:
Change looks good to me, but it would also be good if you can test it (sadly we don't have any end to end tests for integration packs yet).

emedvedev:
Do you want me to try and set up some kind of testing framework? Or should we merge this for now and make testing a separate big issue?

Kami:
Today most of the packs are tested manually and some of them are tested indirectly by the workflows which use those packs.

Getting testing framework set up for all the packs is quite a lot of work (and there are a lot of decision to be made) so ideally you should just quickly manually test it to ensure it still works.

emedvedev:
Test comment, yay!

emedvedev:
Well, it worked. Good to merge? :)

enykeev:
LGTM and Travis is also happy

PR #279, opened 2015-09-01
Fix format error in hubot when posting mistral result

m4dcoder:


packs/hubot/actions/post_result.py
```diff
@@ -37,9 +37,9 @@ def format_possible_failure_result(result):
 
 def format_default_result(result):
     try:
-        output = json.loads(result)
```
manasdk: isinstance(value, six.string_types)

manasdk:
+1

jfryman:
pnda

PR #282, opened 2015-09-03
Feature/octopus pack

tonybaloney:
Created a new pack for Octopus Deploy, Octopus Deploy is one of the most popular DevOps tools for the .NET stack.

The pack includes the key tasks
* create a release of a project
* deploy a release of a project
* get a list of releases for a project

The config file takes the api_key hostname and port of the Octopus API server.


packs/octopusdeploy/actions/deploy_release.yaml
```diff
@@ -0,0 +1,17 @@
...
+runner_type: run-python
+description: Deploy a release in Octopus.
+enabled: true
+entry_point: deploy_release.py
+parameters:
+  release_id:
+    type: string
+    description: ID of the release in Octopus
+    required: true
+    position: 0
```
lakshmi-kannan: Positions are used only for script actions that accept positional arguments. You don't need this for python actions.

packs/octopusdeploy/actions/lib/actions.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def _init_client(self):
+        api_key = self.config['api_key']
+        host = self.config['host']
+        port = self.config['port']
+        return OctopusDeployClient(api_key=api_key, host=host, port=port)
+
+    def _build_uri(self):
+        # big assumption but it'll cover 99% case,
+        # as octopus runs https by default
```
lakshmi-kannan: Can we then set the default to 443? So https by default, http otherwise.

lakshmi-kannan:
<3 it. Just the two comments.

packs/octopusdeploy/actions/create_release.yaml
```diff
@@ -0,0 +1,15 @@
+---
+name: create_release
+runner_type: run-python
```
armab: Seems `run-python` runner type is outdated naming. 
http://docs.stackstorm.com/actions.html?highlight=runner_type#available-runners
It will work, but better to replace with `python-script` for consistency and less confusion in future.

Kami: Yeah, good catch.

We still support old names for backward compatibility reasons, but new code should use `python-script` as you said :)

armab: I'll PR

packs/octopusdeploy/actions/lib/actions.py
```diff
@@ -0,0 +1,59 @@
...
+        start = "http://" if self.client.port is 80 else "https://"
```
Kami: This won't work if a user runs an instance on a non-standard port (80) without SSL, but it's probably not a big deal right now.

I don't know if they even support running it on a non-standard port and SSL should be preferred anyway :)

Kami:
What others have said - nice work :+1:

PR #283, opened 2015-09-03
[octopusdeploy] Replace `run-python` with `python-script` runner_type

armab:
`run-python` runner type is [outdated naming](https://github.com/StackStorm/st2/blob/master/docs/source/upgrade_notes.rst#st2-09).
It will work because of [backward compatibility](https://github.com/StackStorm/st2/blob/badabcb2039dba32fe75e9e5cf7ce03e17f6c350/st2actions/st2actions/bootstrap/runnersregistrar.py#L357), but it's better to replace with `python-script` for consistency and less confusion in future.

http://docs.stackstorm.com/actions.html?highlight=runner_type#available-runners

---

cc @tonybaloney it's really clean & nice pack!

### DIFF
--- a/packs/octopusdeploy/README.md
+++ b/packs/octopusdeploy/README.md
@@ -1,7 +1,8 @@
 # Octopus Deploy Integration Pack
 
 This integration pack allows you to integrate with
-[Octopus Deploy](http://octopusdeploy.com/).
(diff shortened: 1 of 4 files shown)


Kami:
Merged, thanks.

PR #284, opened 2015-09-03
Added create_issue action for GitHub pack

jonico:
Action name: create_issue
Parameters:
* title (string)
* description (string, optional)
* repo (string)
* user (string)
* assignee (string, optional)

Updated README and increased pack version number to 0.2

### DIFF
--- a/packs/github/actions/create_issue.py
+++ b/packs/github/actions/create_issue.py
@@ -1,0 +1,17 @@
+from lib.base import BaseGithubAction
+from github import GithubObject
+from lib.formatters import issue_to_dict
+
+__all__ = [
+    'CreateIssueAction'
+]
+
+
+class CreateIssueAction(BaseGithubAction):
+    def run(self, user, repo, title, description=None, assignee=None):
+        user = self._client.get_user(user)
+        repo = user.get_repo(repo)
+        issue = repo.create_issue(title, description or GithubObject.NotSet,
+                                  assignee or GithubObject.NotSet)
+        result = issue_to_dict(issue=issue)
+        return result
--- a/packs/github/actions/create_issue.yaml
+++ b/packs/github/actions/create_issue.yaml
@@ -1,0 +1,27 @@
+---
+name: create_issue
+runner_type: run-python
+description: Create a Github issue.
+enabled: true
+entry_point: create_issue.py
+parameters:
+  user:
+    type: "string"
+    description: "User / organization name."
+    required: true
+  repo:
+    type: "string"
+    description: "Repository name."
+    required: true
+  title:
+    type: "string"
+    description: Title
+    required: true
+  description:
+    type: "string"
+    description: Description
+    required: false
+  assignee:
+    type: "string"
+    description: Assignee
+    required: false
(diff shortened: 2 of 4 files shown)


packs/github/actions/create_issue.py
```diff
@@ -0,0 +1,18 @@
...
+from github import GithubObject
+from lib.formatters import issue_to_dict
+
+__all__ = [
+    'CreateIssueAction'
+]
+
+
+class CreateIssueAction(BaseGithubAction):
+    def run(self, user, repo, title, description, assignee):
```
jfryman: For any of the optional passed parameters, you'll want to set that parameter to `None`. Otherwise, the runner will get upset when it's expecting that parameter but we're not feeding it through.

packs/github/actions/create_issue.py
```diff
@@ -0,0 +1,18 @@
...
+        #title = int(title)
+
+        user = self._client.get_user(user)
+        repo = user.get_repo(repo)
+	issue = repo.create_issue(title, description or GithubObject.NotSet, assignee or GithubObject.NotSet)
```
lakshmi-kannan: Nit: indentation.

jfryman:
Very nice. Just a few small things, I think, and then we can merge this in and plug in into the community room.

packs/github/actions/create_issue.yaml
```diff
@@ -0,0 +1,27 @@
...
+    description: Title
+    required: true
+  description:
+    type: "string"
+    description: Description
+    required: false
+  assignee:
+    type: "string"
+    description: Assignee
+    required: false
```
lakshmi-kannan: I'd be super happy if we also supported milestone. This is something that'd be useful. Just a request.

packs/github/actions/create_issue.py
```diff
@@ -0,0 +1,18 @@
...
+from lib.formatters import issue_to_dict
+
+__all__ = [
+    'CreateIssueAction'
+]
+
+
+class CreateIssueAction(BaseGithubAction):
+    def run(self, user, repo, title, description, assignee):
+        #title = int(title)
```
jfryman: Is this needed?

packs/github/actions/create_issue.py
jonico: Have you seen my or statements that take care of that?
Or do I mixup something?

packs/github/actions/create_issue.py
jonico: Yeah, flake already caught me, updated

packs/github/actions/create_issue.py
jfryman: Like this: https://github.com/StackStorm/st2contrib/blob/master/packs/mailgun/actions/send_email.py#L15

packs/github/actions/create_issue.yaml
jonico: For milestones and labels it seems as if one has to do additional calls to GitHub's API to get their ids. The GitHub Python library only allows to lookup milestones as ints and to list them all. Would be an expensive operation to always retrieve the list and check whether the param is among them.

packs/github/actions/create_issue.py
jonico: Done

packs/github/actions/create_issue.py
jonico: Line removed

PR #285, opened 2015-09-04
Feature/octopus triggers

tonybaloney:
Created sensors for detecting new releases and new deployments in Octopus.


packs/octopusdeploy/actions/add_machine.yaml
```diff
@@ -0,0 +1,27 @@
...
+  uri:
+    type: string
+    description: Uri of the Octopus tentacle, typically https://host:10933/
+    required: true
+  thumbprint:
+    type: string
+    description: The encryption thumbprint generated by the installer
+    required: true
+  roles:
+    type: string
```
Kami: We supported `array` type so you should use that. This way you don't need to use .split, etc. in the action.

For convenience, CLI allows you to specify `array` parameters as a comma delimited string (those get converted to an array internaly in the CLI) - http://docs.stackstorm.com/cli.html#specifying-parameters-which-type-is-array

packs/octopusdeploy/sensors/new_deployment_sensor.py
```diff
@@ -0,0 +1,136 @@
...
+        self._logger = self._sensor_service.get_logger(__name__)
+        self._deploy_key = 'octopusdeploy.last_deploy_date_str'
+        self._last_date = None
+
+    def setup(self):
+        pass
+
+    def poll(self):
+        self._logger.debug('Requesting list of deployments')
+        deployments = self._get_deployments()
```
Kami: Does API support for kind of filtering (based on a date or similar), or do you always need to retrieve all the available deployments?

tonybaloney:
Sadly not, I tried guessing at the filter parameters, it supports basic paging but not query by date. I will ask the devs on it to extend in the future as it's not ideal. By default it will return 30 results, so at least it's not a huge amount of data

packs/octopusdeploy/sensors/new_release_sensor.py
```diff
@@ -0,0 +1,135 @@
+import time
+
+from lib.sensors import OctopusDeploySensor
+
+
+class NewReleaseSensor(OctopusDeploySensor):
```
Kami: This and the sensor above share a lot of common code. Ideally we could refactor common code into a base class and only override a couple of things which are needed in the sensor specific class (trigger_ref, _to_trigger, datastore key).

Kami:
I've added some comments. Overall, good changes - thanks!

tonybaloney:
I'll refactor the sensors to share a common library.

tonybaloney:
I've refactored the sensor classes, not gone too overboard as once people start using the sensors I think there might need to be additional metadata in the trigger payload. For example at the moment it says a new deployment has been created, but the user will want to know whether it succeeded or failed, which is a seperate API call. And as such, the poll methods are going to become unique.

Kami:
Great, thanks. I will have a look.

Kami:
Sorry for the delay - the changes looks good, but I planned to test them before merging them and I haven't got to that yet.

In any case, since the changes look good I will merge it and test it later.

Kami:
Merged, thanks :)

PR #287, opened 2015-09-08
Update channel -> route

jfryman:
This PR updates the rule in `notify_hubot` to use the updated `routes` vernacular.

### DIFF
--- a/packs/hubot/rules/notify_hubot.yaml
+++ b/packs/hubot/rules/notify_hubot.yaml
@@ -7,7 +7,7 @@ trigger:
   pack: "chatops"
   type: "core.st2.generic.notifytrigger"
 criteria:
-  trigger.routes:
+  trigger.route:
     pattern: "hubot"
     type: "equals"
 action:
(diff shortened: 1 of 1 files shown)


packs/hubot/rules/notify_hubot.yaml
```diff
@@ -7,7 +7,7 @@ trigger:
   pack: "chatops"
   type: "core.st2.generic.notifytrigger"
 criteria:
-  trigger.channel:
+  trigger.routes:
```
Kami: I need to check the code, but shouldn't this be `route`, not `routes`?

If routes is an array / list then equal operator won't work on it.

jfryman: Good question. I'm pretty sure it's singular `route`, since this was nothing more than a rename.

lakshmi-kannan: Singular `route`.

Kami:
LGTM, +1

PR #292, opened 2015-09-16
[trello] TrelloListSensor improvements

armab:
Improves `TrelloListSensor` #291  implementation, - sensor which monitors Trello List for new actions (new Trello card is created, edited, deleted, etc).

* [x] Ability to monitor several Trello Lists
* [x] Every monitored Trello List can use its own API credentials (work with many Trello accounts)
* [x] API credentials precedence: `list config` > `lists config` > `global config` (global config is less significant)
* [x] [Parameters descriptions for `payload_schema`](https://github.com/StackStorm/st2contrib/pull/292/files#diff-ccfcafb9de5144a242c9c39721040ab4R11) (autocomplete hints in webui)
* [x] Better documentation for `TrelloListSensor`
* [x] Monkey (manual) testing

### DIFF
--- a/packs/trello/README.md
+++ b/packs/trello/README.md
@@ -49,20 +49,35 @@ To obtain an oAuth token, refer to the documentation at https://trello.com/docs/
 
 ## Sensors
 ### TrelloListSensor
-Listens for the new Actions in Trello List, which is specified via [config](config.yaml), and dispatches trigger for each new event occurred.
+Listens for the new Actions (changes) in Trello List(s) and dispatches trigger for each new event occurred.
 
-Optionally you can specify in config which actions to listen via `filter` parameter.
+#### Config
+Parameters in `list_actions_sensor` for every trello list:
+* `list_id` - Trello List ID we monitor for new actions (required)
+* `board_id` - Board ID where Trello List is located (required)
+* `api_key` - Trello API key (optional)
+* `token` - Trello API token (optional)
+* `filter` - Filter actions by type(s) (eg. createCard, deleteCard) (optional)
+
 For list of available filters see [Trello API docs](https://trello.com/docs/api/list/index.html#get-1-lists-idlist-actions).
 
-When receives new data, Sensor emits:
+> API credentials work at any level with lower priority for top-level config credentials:
+`list config` > `lists config` > `global config`.
+It means that every Trello list can have its own unique credentials to work with different accounts.
+
+See [config](config.yaml) for more examples.
+Keep in mind, that Sensor with invalid config (bad credentials) will die after 3 connection retries.
+
+#### Output
+When `TrelloListSensor` receives new data, it emits:
 * trigger: `trello.new_action`
-* returns data:
-  * `payload.id` - Action ID (string)
-  * `payload.data` - Main data returned, specific for action, depends on action type (object)
-  * `payload.date` - When action occurred (string)
-  * `payload.idMemberCreator` - User ID who initiated the action (string)
-  * `payload.type` - Action type (ex: createCard) (string)
-  * `payload.memberCreator` - Extended info about user who initiated action (object)
+* and returns data:
+  * `trigger.id` - Action ID (string)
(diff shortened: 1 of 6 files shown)


packs/trello/sensors/list_actions_sensor.yaml
```diff
@@ -9,14 +9,38 @@ trigger_types:
...
           type: string
         memberCreator:
+          description: "Additional info about user who initiated action"
           type: object
-        date:
-          type: string
-          #format: 2015-09-02T18:53:30.900Z
-        id:
-          type: string
+          properties:
```
armab: Nested properties in Sensor `payload_schema` first try,
cc https://github.com/StackStorm/st2web/issues/219

packs/trello/config.yaml
```diff
@@ -3,7 +3,16 @@
...
 
   list_actions_sensor:
-    board_id: c39TEFLt
-    list_id: 55e7456df81e21e9b4aea429
-    filter:
-      - createCard
+    #api_key: ""
+    #token: ""
+    lists:
+    - list_id: 55e7456df81e21e9b4aea429
```
armab: Also still open question: does it makes sense to include Trello list/board IDs as an examples or it's better to wipe everything that looks like real data?
> Since Trello list and Trello board are private - none can access my lists.

I followed [GitHub sensor config](https://github.com/StackStorm/st2contrib/blob/master/packs/github/config.yaml#L7), which lists repositories (and found it easier to understand the Sensor config).

packs/trello/config.yaml
```diff
@@ -3,7 +3,16 @@
   token: ""
 
   list_actions_sensor:
-    board_id: c39TEFLt
-    list_id: 55e7456df81e21e9b4aea429
-    filter:
-      - createCard
+    #api_key: ""
+    #token: ""
+    lists:
```
armab: Unclear question about monitoring several GitHub repositories/several independent Trello Lists in one Sensor with big shared config: 
https://github.com/StackStorm/st2/pull/1739#issuecomment-140725275

packs/trello/README.md
```diff
@@ -47,6 +47,39 @@ To obtain an oAuth token, refer to the documentation at https://trello.com/docs/
...
+It means that every Trello list can have its own unique credentials to work with different accounts.
+
+See [config](config.yaml) for more examples.
+Keep in mind, that Sensor with invalid config (bad credentials) will die after 3 connection retries.
+
+#### Output
+When `TrelloListSensor` receives new data, it emits:
+* trigger: `trello.new_action`
+* and returns data:
+  * `trigger.id` - Action ID (string)
```
Kami: Is the list id also included in the payload?

Since we support monitoring multiple lists I think it should be, but that's not visible in the code because we directly dispatch the payload as returned by the API.

Kami: An example sensor trigger payload, similar to what we have for other sensors would also be nice (e.g. https://github.com/StackStorm/st2contrib/tree/master/packs/cubesensors#sensors).

armab: Yes, `list.id` is part of `trigger.data`. [Example from API](https://developers.trello.com/advanced-reference/list#get-1-lists-idlist-actions): 

```json
{
    "id": "4efe3147c72846af4e00006d",
    "data": {
        "list": {
            "id": "4eea4ffc91e31d174600004a",
            "name": "To Do Later"
        },
        "board": {
            "id": "4eea4ffc91e31d1746000046",
            "name": "Example Board"
        },
        "old": {
            "name": "To Do Eventually"
        }
    },
    "date": "2011-12-30T21:46:47.843Z",
    "idMemberCreator": "4ee7deffe582acdec80000ac",
    "type": "updateList",
    "memberCreator": {
        "id": "4ee7deffe582acdec80000ac",
        "avatarHash": null,
        "fullName": "Joe Tester",
        "initials": "JT",
        "username": "joetester"
    }
}
```

Kami: Thanks.

Kami:
Merged, thanks!

PR #293, opened 2015-09-24
Add new xml pack

Kami:
It includes an initial `parse` action which parses XML string and returns JSON object.

This action allows users to work with XML documents / data (ingest or work with them inside a workflow) in StackStorm.


armab:
Rocks :100:

jfryman:
👍

giampierod:
Woot. Thanks.

PR #298, opened 2015-10-05
Prepend ansible pack virtualenv path to PATH env variable

### LINKED ISSUE #296: Ansible pack won't install binaries when ansible is already installed on the machine
I had ansible already on the machine, installed via apt-get and when I added the ansible pack for stackstorm, it didn't add the binaries.
At the beginning I had my other ansible machine just symlink the binaries but after removing ansible on the stackstorm machine, the pack installed the binaries correcty.

armab:
This way if ansible is not found in st2 pack virtualenv directory, global ansible will be used as a fallback (tested).
Fixes #296

### DIFF
--- a/packs/ansible/actions/lib/ansible_base.py
+++ b/packs/ansible/actions/lib/ansible_base.py
@@ -21,6 +21,19 @@ class AnsibleBaseRunner(object):
         :type args: ``list``
         """
         self.args = args[1:]
+        self._prepend_venv_path()
+
+    @staticmethod
+    def _prepend_venv_path():
+        """
+        Modify PATH env variable by prepending virtualenv path with ansible binaries.
+        This way venv ansible has precedence over globally installed ansible.
+        """
+        venv_path = '/opt/stackstorm/virtualenvs/ansible/bin'
+        old_path = os.environ.get('PATH', '').split(':')
+        new_path = [venv_path] + old_path
+
+        os.environ['PATH'] = ':'.join(new_path)
 
     def execute(self):
         """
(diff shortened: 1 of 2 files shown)


jfryman:
LGTM

packs/ansible/actions/lib/ansible_base.py
```diff
@@ -21,6 +21,19 @@ def __init__(self, args):
...
         self.args = args[1:]
+        self._prepend_venv_path()
+
+    @staticmethod
+    def _prepend_venv_path():
+        """
+        Modify PATH env variable by prepending virtualenv path with ansible binaries.
+        This way venv ansible has precedence over globally installed ansible.
+        """
+        venv_path = '/opt/stackstorm/virtualenvs/ansible/bin'
```
Kami: To make it more robust can you please use:

```python
base_virtualenvs_path = os.path.join(cfg.CONF.system.base_path,  'virtualenvs/')
...
```

Kami:
Thanks for making this change :)

packs/ansible/actions/lib/ansible_base.py
armab: I was looking for something like that when initially implemented ansible pack, to avoid this `/opt/stackstorm/virtualenvs/ansible/bin` hardcoding string.

But `cfg.CONF.system` is available only if `python-script` is used (part of [sandboxing](https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/sandboxing.py)), not for standalone `local-shell-script`. No?

Kami: Ah, sorry, for some reason I keep forgetting it's not a Python runner action, heh :)

armab: Yeah np,
don't like that `/opt/stackstorm/virtualenvs/ansible/bin` too.

PR #299, opened 2015-10-07
Added extra actions (get deployments and get projects) to the pack an…

tonybaloney:
Additional actions to the pack

### DIFF
--- a/packs/octopusdeploy/actions/get_deployments.py
+++ b/packs/octopusdeploy/actions/get_deployments.py
@@ -7,6 +7,8 @@ __all__ = [
 
 class GetDeploymentsAction(OctopusDeployAction):
     def run(self, project_id):
-        result = self.make_get_request(action="deployments?projects="
-                                       % project_id)
-        return result['Items']
+        result = self.make_get_request(action="deployments",
+                                       params={
+                                           'projects':project_id
+                                       })
+        return result.get('Items', None)
--- a/packs/octopusdeploy/actions/get_deployments.py
+++ b/packs/octopusdeploy/actions/get_deployments.py
@@ -9,6 +9,6 @@ class GetDeploymentsAction(OctopusDeployAction):
     def run(self, project_id):
         result = self.make_get_request(action="deployments",
                                        params={
-                                           'projects':project_id
+                                           'projects': project_id
                                        })
         return result.get('Items', None)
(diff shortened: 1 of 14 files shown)


packs/octopusdeploy/actions/get_deployments.py
```diff
@@ -0,0 +1,12 @@
...
+
+__all__ = [
+    'GetDeploymentsAction'
+]
+
+
+class GetDeploymentsAction(OctopusDeployAction):
+    def run(self, project_id):
+        result = self.make_get_request(action="deployments?projects="
+                                       % project_id)
```
lakshmi-kannan: Would be better to expose `params` and `headers` as argument as use them instead of handcrafting the query params.
https://github.com/StackStorm/st2contrib/blob/master/packs/octopusdeploy/actions/lib/actions.py#L46

packs/octopusdeploy/actions/get_deployments.py
```diff
@@ -0,0 +1,12 @@
...
+        return result['Items']
```
lakshmi-kannan: Just noting that irrespective of what status code the HTTP request produces this action will always pass. If you want the action to fail on non 2XX code, you'll need the base method to return status and throw an exception to fail the action. 

Also, I highly recommend using .get('Items', None) :P.

tonybaloney: Which lib are the St2 error types defined?

packs/octopusdeploy/actions/get_deployments.py
```diff
@@ -0,0 +1,14 @@
...
+]
+
+
+class GetDeploymentsAction(OctopusDeployAction):
+    def run(self, project_id):
+        result = self.make_get_request(action="deployments",
+                                       params={
+                                           'projects': project_id
+                                       })
+        return result.get('Items', None)
```
Kami: It looks like this API call returns a list. If so it might be better to return empty list instead if the `Items` attribute is not available.

This way it's easier to use it inside the actions since you don't need to check for None.

packs/octopusdeploy/actions/get_deployments.yaml
```diff
@@ -0,0 +1,11 @@
+---
+name: get_deployments
```
Kami: Minor style / consistency thing - we tend prefix actions which return a list with `list_` (e.g. `list_deployments`) and methods which return a single item with `get_` (e.g. `get_deployment`).

Kami:
Thanks - I've added some comments.

Some small style / consistency stuff. Besides that, LGTM.

tonybaloney:
Done and done

tonybaloney:
all fixed

Kami:
Merged, thanks.

PR #302, opened 2015-10-25
Collect anonymous data action fixes

manasdk:
* Previous the checks would not honor the flags so reworking that to handle the flags correctly.


manasdk:
Validate `st2.call_home` with both `st2::collect_anonymous_data` set and unset.

manasdk:
Merging.

PR #306, opened 2015-11-02
Fix rule names to not include packs

lakshmi-kannan:


packs/hubot/rules/notify_hubot.yaml
```diff
@@ -1,6 +1,7 @@
 # notify_hubot.yaml
 ---
-name: "chatops.notify_hubot"
```
lakshmi-kannan: I am not sure if this change will break our chatops. I'll validate.

lakshmi-kannan: It shouldn't. So this is fine.

PR #307, opened 2015-11-03
Update st2.upload_to_s3 action so it can upload to any publicly writable S3 bucket, remove AWS credentials from config

Kami:
This pull request updates `st2.upload_to_s3` action so it supports uploading to any publicly writable S3 bucket. This way we don't need to store any expose S3 credentials in the repo (credentials had limited permissions and only allowed user to write to a specific bucket, but it's still a bad practice).

We use same approach for uploading debugging information to S3. For that to work, we need to configurable bucket's IAM policies so everyone can write to it and only we can read from it (I've already done that).

### DIFF
--- a/packs/st2/actions/upload_to_s3.py
+++ b/packs/st2/actions/upload_to_s3.py
@@ -1,21 +1,51 @@
+import os
+import httplib
+
+import requests
+
 from lib.action import St2BaseAction
-import s3
 
 __all__ = [
     'UploadToS3'
 ]
 
+S3_BUCKET_URL = 'https://%(bucket)s.s3.amazonaws.com/'
+
 
 class UploadToS3(St2BaseAction):
-    def run(self, file_name, remote_file, bucket):
-        remote_name = s3.S3Name(remote_file, bucket=bucket)
+    """
+    Note: The destination bucket me be configured so everyone can write to it,
+    but only we can read from it.
+    """
 
-        connection = s3.S3Connection(**self.config['s3'])
-        storage = s3.Storage(connection)
-        storage.write(file_name, remote_name)
-        payload = {
-            "status": "ok",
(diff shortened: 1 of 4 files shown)


packs/st2/actions/upload_to_s3.py
```diff
@@ -1,21 +1,51 @@
...
-        remote_name = s3.S3Name(remote_file, bucket=bucket)
-
-        connection = s3.S3Connection(**self.config['s3'])
-        storage = s3.Storage(connection)
-        storage.write(file_name, remote_name)
-        payload = {
-            "status": "ok",
-            "uploaded_file": remote_file,
+    """
+    Note: The destination bucket me be configured so everyone can write to it,
```
lakshmi-kannan: me be? fat fingers :P

packs/st2/actions/upload_to_s3.py
```diff
@@ -1,21 +1,51 @@
...
+
+        if not os.path.isfile(file_name):
+            raise ValueError('Local file "%s" doesn\'t exist' % (file_name))
+
+        url = S3_BUCKET_URL % {'bucket': bucket}
+        url = url + remote_file
+
+        # Note: Requests library performs streaming and chunked upload
+        files = {'file': open(file_name, 'rb')}
+        response = requests.put(url=url, files=files, verify=False)
```
lakshmi-kannan: Why are you using requests instead of s3 client library? Just to avoid deps?

lakshmi-kannan: verify=False may not be desired?

Kami: There is no need to use S3 library here since it's a simple PUT http request.

I also removed verify=False, I used it for debugging purposes.

PR #309, opened 2015-11-04
Fixing pack name in a hubot rule

emedvedev:


Kami:
LGTM

packs/hubot/rules/notify_hubot.yaml
```diff
@@ -1,11 +1,11 @@
 # notify_hubot.yaml
 ---
 name: "notify_hubot"
-pack: "chatops"
+pack: "hubot"
 enabled: true
 description: "Notification rule to send messages to Hubot"
 trigger:
-  pack: "chatops"
+  pack: "hubot"
```
Kami: I think this attribute is invalid here anyway since trigger it's actually in the core pack and you can remove it all together.

PR #310, opened 2015-11-04
Rewrite hubot actions to use stream API

enykeev:


packs/hubot/actions/format_result.py
```diff
@@ -139,37 +139,9 @@ def format_windowsrunner_result(result):
...
-        data = json.dumps(body)
-        self.logger.info(data)
-        response = requests.post(url=url, headers=headers, data=data)
-
-        if response.status_code == httplib.OK:
-            self.logger.info('Message successfully posted')
-        else:
-            self.logger.exception('Failed to post message: %s' % (response.text))
-        return True
+class FormatResultAction(Action):
```
manasdk: How about removing all the other lines of code?

packs/hubot/actions/workflows/post_result.yaml
```diff
@@ -0,0 +1,18 @@
+---
+  chain:
+    -
+      name: "format_result"
```
manasdk: I thought the main motivation here was to do no formatting in python and move it all to hubot. I feel we should simply post the entire execution or payload of the notification.

packs/hubot/rules/notify_hubot.yaml
```diff
@@ -1,10 +1,10 @@
 # notify_hubot.yaml
 ---
-name: "notify_hubot"
-pack: "hubot"
+name: "chatops.notify_hubot"
```
manasdk: This looks wrong.

name : notify_hubot
pack : hubot

packs/hubot/rules/notify_hubot.yaml
```diff
@@ -1,10 +1,10 @@
...
 enabled: true
 description: "Notification rule to send messages to Hubot"
 trigger:
+  pack: "chatops"
```
manasdk: not required.

packs/hubot/rules/notify_hubot.yaml
enykeev: oh, sorry, merging issue

packs/hubot/actions/workflows/post_result.yaml
enykeev: The main motivation for this particular change was to get rid of http call to hubot. I've decided not to touch formatting part partly because a) converting the results based on runner type is not necessarily a thing we want clients to implement and I thought it was you who proposed we keep this kind of logic on st2 side, b) because I've decided to solve one problem at a time and c) formatting is the thing @emedvedev is working on.

I'm not intending to merge this PR just now, I just wanted to give you an impression of how it may look like. If you want formatting happen on bot side, sure, we'll sync with @emedvedev and see where our two PRs would meet.

enykeev:
@StackStorm/team The real question here is what do we do to prevent the pack updating on installations prior to 1.2?

packs/hubot/actions/format_result.py
enykeev: Is it better now?

packs/chatops/actions/templates/default.j2
```diff
@@ -0,0 +1,44 @@
+{%- macro serialize(iter, prefix='') -%}
```
manasdk: What the heck is this file?

enykeev: Jinja template? Because fitting something like that into the python code and then struggling with formatting is not fun =)

enykeev:
Temporarily merged to StackStorm/st2incubator#280 until we figure out whether this is how we want to implement it.

emedvedev:
Closing because we maintain this pack in the core now.

PR #311, opened 2015-11-10
Merging in auto generated slack actions

DoriftoShoes:
This PR merges in the actions from the auto-generated Slack pack:

All existing actions and configs remain for backwards compatibility.


manasdk:
Are we abandoning st2-slack?

DoriftoShoes:
We need to collapse these one way or the other.  I went this way so people that already had the pack can just reinstall and get the updates...no need to find the new location.

PR #312, opened 2015-11-12
Allow picking fields to exclude for executions.get

enykeev:


packs/st2/actions/lib/action.py
```diff
@@ -47,7 +47,7 @@ def _get_auth_token(self):
 
         return token
 
-    def _run_client_method(self, method, method_kwargs, format_func):
+    def _run_client_method(self, method, method_kwargs, format_func, format_kwargs={}):
```
Kami: Should default to a non mutable type (e.g. None and do format_kwargs = format_kwargs or {} in the body)/

Kami:
LGTM, +1

packs/st2/actions/lib/action.py
manasdk: Like Kami said - 
https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

enykeev: yeah, forgot about that. Thanks!

PR #313, opened 2015-11-12
Refactored pack to use a consistent formatter and field sets, support…

tonybaloney:
Refactored pack to use a consistent formatter and field sets, supports lists and single instances, also supports 1-level deep field mapping to the dictionary. Also added new actions, list_images, list_sizes.

### DIFF
--- a/packs/libcloud/actions/balancer_attach_member.py
+++ b/packs/libcloud/actions/balancer_attach_member.py
@@ -6,10 +6,6 @@ __all__ = [
     'BalancerAttachMemberAction'
 ]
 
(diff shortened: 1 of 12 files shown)


manasdk:
I bet Kami has strong views on this one. 

Overall, I see the direction and yeah it does make sense.  If we could up with a `generic` serializer that would be even better but then often it is better to remain pragmatic.

Kami:
Thanks and sorry for the delay.

I will look into this today.

Kami:
Good change. LGTM +1.

I will merge this and push some minor style changes afterwards.

PR #319, opened 2015-11-25
Allow non dictionary based queries in the ServiceNow pack

tonybaloney:


packs/servicenow/actions/get_non_structured.py
```diff
@@ -0,0 +1,8 @@
+from lib.actions import BaseAction
+
+
+class GetNonStructuredAction(BaseAction):
+    def run(self, table, query):
+        self.client.table = table  # pylint: disable=no-member
+        response = self.client.get(query)  # pylint: disable=no-member
+        return response
```
lakshmi-kannan: Can you validate response actually returns a json string? I am not sure if you need response.json().

Kami: Looks like it returns a parsed JSON response - https://github.com/rbw0/python-servicenow-rest/blob/master/servicenow_rest/api.py#L81

Kami:
LGTM, +1

PR #324, opened 2015-11-27
changed to new webhook format

sachinpk:
The newrelic webhook payload has changed so old newrelic pack doesn't works ,It ignores incomming webhook. I've converted this sensor to accept new webhooks

the format of new webhook is like given below

{
"current_state":"open", <--[open|acknowledged|closed]
"details":"Apdex < .80 for at least 10 min",
"severity":"WARN", <--[INFO|WARN|CRITICAL]
"incident_api_url":"https://alerts.newrelic.com/api/explore/applications/incidents/1234",
"incident_url":"https://alerts.newrelic.com/accounts/99999999999/incidents/1234",
"owner":"user name",
"policy_url":"https://alerts.newrelic.com/accounts/99999999999/policies/456",
"runbook_url":"https://localhost/runbook",
"policy_name":"APM Apdex policy",
"condition_id":987654,
"condition_name":"My APM Apdex condition name",
"event_type":"INCIDENT",
"incident_id":1234,
"targets":[
{
"id":"23456",
"name":"My Application",
"link":"https://alerts.newrelic.com/accounts/99999999999/applications/23456?tw[start]=1425415729&tw[end]=1425417529"
"labels":{
"Service":"Account Service",
"Product":"APM",
"Team":"APM Team",
},
"type":"Application",
}
]
"timestamp":1425417514074
}

PR #325, opened 2015-11-30
Add back sensor for the New Relic legacy API and alerting

Kami:
#324 updated New Relic sensor to work with the new API. This pull request adds back an old sensor which works with the legacy API - it's quite likely that some people still use the legacy API and alerting.

To avoid port conflicts, only the new sensor is enabled by default and if the user wants to use the old one, they need to enable old one and disable the new one.


Kami:
I've added some basic docs on the sensors to the README, but we still need some better docs which explain in more detail how the sensor works (it exposes an HTTP server for receiving webhooks, etc.).

Kami:
It looks like pylint step is failed - it seems like it's using latest version of pylint which breaks a lot of things, but I'm not really sure how this is happening. pylint is pinned to an old version in requirements-dev.txt

Kami:
I'll go ahead and merge it - I can't figure out why pylint is failing (it could be corrupted pip cache on Travis). I can't reproduce failure locally even with pylint 1.5.0.

Kami:
Looks like astroid dependency was causing failures - https://github.com/StackStorm/st2/pull/2247

Will merge the fix into this repo as well.

PR #326, opened 2015-11-30
re:Amaze Integration Pack

jfryman:
This PR introduces the basic skeleton for the `re:Amaze` SaaS KB/Ticket Tracker tool. Specifically focusing on getting search values from the KB at this time.


packs/reamaze/actions/article_search.py
```diff
@@ -0,0 +1,32 @@
+from lib.actions import BaseAction
+from dotted.utils import dot
+from fuzzywuzzy import fuzz
+import json
+
+
+class ArticleSearch(BaseAction):
+    def run(self, query, include_body=False, limit=5):
+        response = self._api_get('/articles')
```
manasdk: is there some pagination support here?

jfryman: Nope, not today.

jfryman: https://www.reamaze.com/api/get_article

manasdk: At some point this could return a large number of articles. I guess we better watch for native search or pagination support.

jfryman: Indeed. I've already reached out to them to try and figure this out long-term. This gets us short-term

lakshmi-kannan:
Yep, other than the pagination and CPU cycles this thing is going to take, lgtm.

packs/reamaze/actions/article_search.py
```diff
@@ -0,0 +1,45 @@
...
+                               include_body=include_body)
+        response = map(response_fun, results)
+        return json.dumps(response)
+
+    # Takes an incoming query and an article, and runs a fuzzy token
+    # search to get data from an article.
+    #
+    # Returns a tuple of (rank, article)
+    def _rank_article(self, article, query):
+        rank = fuzz.token_set_ratio(query, article['body'].encode('utf-8'))
```
jfryman: I don't like the fact that I have to `encode()` here, but otherwise it fails. I tried `r.encode('utf-8')` in `lib/actions.py#_apy_get`, but that did not do it. Thoughts?

lakshmi-kannan: it fails == what? fuzz library?

manasdk: My guess is that if the `content-type` was `application/json; charset=utf8` the encoding would be setup correctly. 

If you set `r.encoding = utf-8` before `r.json()` does that work better?

jfryman: @manasdk I can try the content type explicitly, and see what happens.

I set `r.encoding` before `r.json()` previously, and no :game_die:

jfryman: @lakshmi-kannan yeah, the fuzz library fails. I get an ASCII parse error.

jfryman: Anyway, not going to block on this. I'll make a note in code for future us to understand and hate current/past us for.

jfryman:
Well, look at this. re:Amaze already helped us out and updated docs. There *is* a query param. :)

Much simpler search, and no more overhead with at least search. Pagination may be an issue, but let's save that for another day.

manasdk:
> Well, look at this. re:Amaze already helped us out and updated docs. There is a query param. :)

nice!

PR #328, opened 2015-12-02
Work around hubot-help's kinks

enykeev:
`hubot-help` simply replaces any occurrence of 'hubot' with either bot's alias or name. This results in rather strange looking lines like `! ! branch on {{hosts}} - Show the currently deployed ! branch`. By very least, it breaks an immersion and makes our bot looks worse than it is.

### DIFF
--- a/packs/hubot/README.md
+++ b/packs/hubot/README.md
@@ -17,6 +17,6 @@ Pack that provides management/integration with Hubot
 
 ## ChatOps Aliases
 
-* `hubot branch`            -> hubot.branch
-* `hubot deploy {{branch}}` -> hubot.deploy
-* `hubot restart`           -> hubot.restart
+* `bot branch`            -> hubot.branch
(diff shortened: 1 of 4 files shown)


enykeev:
With hubot deployed inside docker containers, this pack as a whole makes very little sense, but still, this seems like a problem worth fixing.

Kami:
LGTM

emedvedev:
Yeah, that's what I was going to ask: isn't this pack going to be deprecated anyway?

enykeev:
May be at some point in time when we ensure there's no more standalone hubot deployments left in a wild. On the other hand, we may want to rewrite it to work with docker... if there's actually enough value it it.

jfryman:
:+1: 

Make the experience consistent. We can re-eval long-term.

emedvedev:
We merged a fix into `hubot-help` instead, closing.

PR #331, opened 2015-12-08
re:amaze: Adding ability to post to conversation

jfryman:
This PR adds the ability to post to a conversation in re:amaze from StackStorm. Supports all options/parameters except the ability to impersonate another user.


emedvedev:
@jfryman is the man! Definitely need to have an alias for this though. :P

PR #336, opened 2015-12-10
Update slack sensor trigger `ts` from int to string

jfryman:
This PR updates the `slack` sensor, changing the data type of the `ts` return type from an Integer to a String.

I found this while trying to figure out how to dereference a message to a permalink. Here's how the story goes.

A permalink looks like this: https://stackstorm.slack.com/archives/bot-testing/p1449711449000413

And the accompanying payload that comes along with it, looks like this. 

```
{
    "type": "message",
    "channel": "C2147483705",
    "user": "U2147483697",
    "text": "<https://github.com/StackStorm/st2puppet/pull/196> test 5",
    "ts": "1449711449.000413"
}
```

I've been trying to figure out how to deref a message to a permalink, and been unsuccessful. This is because what I *actually* get something like this...

```
{
    "type": "message",
    "channel": "C2147483705",
    "user": "U2147483697",
    "text": "<https://github.com/StackStorm/st2puppet/pull/196> test 5",
    "ts": 1449711449
}
```

`ts` is cast as an integer, and I lose the resolution necessary to derive a permalink. From what I can tell in the API (https://api.slack.com/events/message), this is a string, so I'm going with it.


manasdk:
Seems reasonable

+1

jfryman:
Well, this isn't the fix, but it still needs to happen. Going to have to dig a bit...

PR #337, opened 2015-12-10
Slack sensor: maintain timestamp resolution

jfryman:
A rework of #336... Brought in...

This PR updates the `slack` sensor, changing the data type of the `ts` return type from an Integer to a String.

I found this while trying to figure out how to dereference a message to a permalink. Here's how the story goes.

A permalink looks like this: https://stackstorm.slack.com/archives/bot-testing/p1449711449000413

And the accompanying payload that comes along with it, looks like this. 

```
{
    "type": "message",
    "channel": "C2147483705",
    "user": "U2147483697",
    "text": "<https://github.com/StackStorm/st2puppet/pull/196> test 5",
    "ts": "1449711449.000413"
}
```

I've been trying to figure out how to deref a message to a permalink, and been unsuccessful. This is because what I *actually* get something like this...

```
{
    "type": "message",
    "channel": "C2147483705",
    "user": "U2147483697",
    "text": "<https://github.com/StackStorm/st2puppet/pull/196> test 5",
    "ts": 1449711449
}
```

`ts` is cast as an integer, and I lose the resolution necessary to derive a permalink. From what I can tell in the API (https://api.slack.com/events/message), this is a string, so I'm going with it.


jfryman:
Neither is this, apparently. I'm going to have to unpack this a bit. Something not quite obvious is happening, and I need :zzz:.

packs/slack/sensors/slack_sensor.py
```diff
@@ -117,7 +118,7 @@ def _handle_result(self, result):
 
         for item in result:
             item_type = item['type']
-            item_timestamp = int(float(item.get('ts', 0)))
+            item_timestamp = int(item.get('ts', 0))
```
Kami: This will throw if `ts` attribute is float (string with a dot in it).

packs/slack/sensors/slack_sensor.py
```diff
@@ -179,7 +180,8 @@ def _handle_message(self, data):
                 'topic': channel_info['topic']['value'],
                 'is_group': channel_info.get('is_group', False),
             },
-            'timestamp': int(float(data['ts'])),
+            'timestamp': int(string.replace(data['ts'], '.', ''))),
```
Kami: So this value is now representing microseconds since the epoch?

If so, we should probably add a new attribute since timestamp usually refers to a unix epoch in seconds.

PR #338, opened 2015-12-10
Add raw Timestamp to Slack Payload

jfryman:
For the immediate need.

There is some mangling happening, and I need the raw string. (See #336 or #337). So, instead of trying to fix the thing, here is an addition. Less ideal, but unblocks me and I can get going. 

Fixes appreciated. :heart:

### DIFF
--- a/packs/slack/sensors/slack_sensor.py
+++ b/packs/slack/sensors/slack_sensor.py
@@ -180,6 +180,7 @@ class SlackSensor(PollingSensor):
                 'is_group': channel_info.get('is_group', False),
             },
             'timestamp': int(float(data['ts'])),
+            'timestamp_raw': data['ts'],
             'text': text
         }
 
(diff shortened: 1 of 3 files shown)


packs/slack/sensors/slack_sensor.py
```diff
@@ -180,6 +180,7 @@ def _handle_message(self, data):
                 'is_group': channel_info.get('is_group', False),
             },
             'timestamp': int(float(data['ts'])),
+            'timestamp_raw': data['ts'],
```
Kami: Yeah, adding additional attribute is a better since it's good to still preserve timestamp as an int so user perform different comparison operations on it.

Kami: On a somewhat related note - there is also a `number` type in json schema which represents numbers which can also hold floats.

This might come handy in some cases.

PR #340, opened 2015-12-11
Ansible underscores

emedvedev:
Since we shouldn't allow `-` in parameters (even if it works sometimes), I'm rewriting the Ansible pack to use underscores and replace them with dashes only when the command is executed. This saves us a lot of trouble (like https://github.com/StackStorm/st2/issues/2280), and also, well, we should follow our own spec. :)

Potentially it could break some old tutorials and examples, but as a user you only need to take a brief look at the pack to understand what has to be changed to make the tutorial work, so I say let's roll with it.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1449838664012195

armab:
As discussed in Slack, I'm :-1: 100500 to avoid this, **if possible**.

Because parameter names in Ansible pack 100% exactly the same as they are in Ansible itself:
```
# ansible pack
ansible.playbook playbook=playbook.yml inventory-file=hosts.ini

# original ansible equivalent:
ansible-playbook playbook=playbook.yml --inventory-file=hosts.ini
```
It's about consistency with "upstream" params and ease of use, so potential users work with the same parameters as they worked with Ansible before and don't need to remember any new naming or custom rules. Ansible is intented to be simple as possible and usage simple as possible, that's probably the reason behind its popularity. Same approach was chosen for integration pack.

For me it looks like global limitation when recent change breaks all parameters with `-` in name. At a high level CLI param naming: `param_name_only_underscores_allowed=value` looks like downgrade and it's not only `ansible` pack problem.
Probably other users will have this with custom packs, since we started re-using param names somewhere as Python variables with recent ChatOps change.

-----

As a conclusion, at a high-level I'm against this, at a low-level I understand all the reasons behind the change.
If we follow this, make sure to do strict validation for action parameter naming globally: stackstorm/st2#2280 So we'd say via validation: `dashes prohibited` for all action parameter names from now or even for all keys names in any yaml meta, since they potentially could be re-used as variables somewhere.

My question: Is it's possible to hack/avoid somehow this special case, with kind of pre-processing/replacing on `st2` side and how dirty it will be (like shooting in own foot)?
Any ideas @StackStorm/team ?

packs/ansible/actions/lib/shell.py
```diff
@@ -17,9 +17,14 @@ def replace_args(attribute=None):
     def _replace_args(f):
         @functools.wraps(f)
         def _wrapper(self, *args):
+            def _replace(arg):
+                for rule in rules:
+                    if arg.startswith(rule):
```
armab: Won't `arg.startswith` mess these rules: 
```
'--verbose=vv': '-vv',
'--verbose=vvv': '-vvv',
'--verbose=vvvv': '-vvvv', 
```
when `--verbose=vvvv` in fact starts with `--verbose=vvv` and `--verbose=vv`?

emedvedev: I changed the order so that `'--verbose=vvvv': '-vvvv'` would resolve first and not propagate to the "lesser" settings.

emedvedev:
One other note: it's not new, the first time I've seen a report about this was a ~month ago.

emedvedev:
That said, I agree with all the rest, and I'm open to any approach where we could avoid this change. I just don't really see it.

manasdk:
> My question: Is it's possible to hack/avoid somehow this special case, with kind of pre-processing/replacing on st2 side and how dirty it will be (like shooting in own foot)?

This is really a jinja limitation. I think there are other things that blew up as well. 

Can we hack?
Yes.

Should we hack?
Not yet. This will lead to much obscure code in parameter handling and I would much rather stay away from having to worry about that right now.

>  make sure to do strict validation for action parameter naming globally

I feel we get off the hook with conventions and docs. Something better than https://docs.stackstorm.com/actionchain.html?highlight=dashes#gotchas. We could perform strict validation but again trying to reduce the problem to know limitation - we will likely fix it soon but bear with us in the meanwhile.

Let replace the the `-` with `_` since it is by far the safer option. Yes this leads to inconsistency with existing CLI but you know we can continue to paying that price. 

We can put `-` support on the backlog and prioritize work on this early next year. (Said for dramatic effect its only really 3 weeks away)

emedvedev:
That was my reasoning as well. Let's get another +1 from someone, and if there's really no better and/or safer option, go ahead and merge it (or I will do it on Monday).

Kami:
I remember there were some issues with dashes in the past, but I don't actually remember where (might have been in the pack names).

It seems that right now the issue only exists in the new ChatOps implementation which is regex based (IIRC, it worked in the past). If that's indeed the only place where it breaks (please correct me if there are more places), I think we should fix this in the ChatOps implementation.

As far as parameter name validation goes - I do agree that we should do a strict validation, the same as we do for pack name (we should first define valid characters, right now I believe we allow a-z-_0-9). In fact we might already do that, I believe I worked on a similar change in the past.

emedvedev:
> IIRC, it worked in the past
It didn't, the issue is not new. This report, for example, predates the new chatops implementation: https://github.com/StackStorm/hubot-stackstorm/issues/52
We also strongly discourage dashes in the docs, as @manasdk pointed out:
https://docs.stackstorm.com/actionchain.html?highlight=dashes#gotchas

Kami:
Also another thing - since as @emedvedev pointed out (https://github.com/StackStorm/st2/issues/2280#issuecomment-163835201), we use named groups in new ChatOps parser implementation and this kinda means we pushed ourselves in a corner.

Named group names must be valid Python identifiers and that's not true for our parameter names (we allow more broader set of characters). For example, `-` is not only offender, we also allow `.` in the parameter names, but `.` is not a valid identifier for a Python variable.

In short - we have a bigger problem and we need to decide how to proceed (this quick "hack" won't solve the root cause) and fix the root cause. It seems like we have two options:

1. Modify parameter name validation and change which characters are valid / allow
2. Fix / change ChatOps parsing implementation

I'm leaning towards the second option since it seems like the right thing - it's fixing a small implementation detail while the first change is a bigger change and it's changing validation because of an implementation limitation (which doesn't seem right, imo, unless there are also other good reasons for limiting valid characters).

emedvedev:
> we also allow . in the parameter names

Because nothing could possibly go wrong with this.

Actually, if we do allow a broader set, fixing the implementation is not such a bad idea, but I'd like to call out allowing `.` and `-` in param names as questionable. What's the complete set?

Kami:
Actually, I just checked the implementation and it seems we use `\w` for parameter name validation. That means `[A-Za-z0-9_].` so `.` are not allow, my bad.

In general, I agree that not allowing `-` in parameter names is a good idea. Having said that, the fix should also include:

1. Updating documentation, making it clear which characters are allowed in pack and action parameter names (we do allow `.` in pack names and if we won't allow it in action names we need to point this out)
2. Update parameter name validation code - we should fail loudly and not load the action if the parameter name validation fails

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450119166013095

armab:
Returning to this case:
I still think it's no good from high-level/user's side in this special case.

For example here https://github.com/StackStorm/st2/issues/2280#issuecomment-163835201 we could hack: kind of replacing dashes with unicode equivalent and vice-versa in several place to avoid regex parsing errors. And Jinja templating is not the case, since Ansible pack is immutable thing and we don't reuse "wrong" variables like `{{ inventory-file }}` there.

From my understanding it's the way of fixing the problem in programming way (and yeah, it's dirty hack even I hate), instead of limiting user's feature.

----

Anyways:
we have `3 x` :+1: 
and `1 x` :-1: (from me).

@emedvedev let's merge this PR and move forward.
I'll update later all Ansible tutorials/showcases to reflect this change, including new ChatOps features.

emedvedev:
Ok, merging, although somewhat reluctantly. Let me know if you need help with new features!

PR #342, opened 2015-12-14
Add makefile target which tries to register all the pack resources

Kami:
This pull request adds a new make target and script which tries to register all the pack resources and exits with non-zero if registration fails.

This will allow us to detect all kind of resource metadata related issues early on (invalid syntax in the metadata file, invalid parameter name, etc.).

Depends on https://github.com/StackStorm/st2/pull/2291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,9 @@ env:
     - "TASK=packs-resource-register"
     - "TASK=packs-tests"
 services:
-  # Note: MongoDB is needed for packs-resource-register check
+  # Note: MongoDB and RabbitMQ is needed for packs-resource-register check
   - mongodb
+  - rabbitmq
 cache:
   directories:
     - $HOME/.pip-cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ services:
   # Note: MongoDB and RabbitMQ is needed for packs-resource-register check
   - mongodb
   - rabbitmq
+addons:
+  apt:
+    packages:
+      - libsox-dev  # needed by witai pack
 cache:
   directories:
     - $HOME/.pip-cache/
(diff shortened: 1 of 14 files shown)


Kami:
Once this is merged, I will also cherry-pick some changes from #332 - this is one of the standalone tasks which doesn't depend on other files so we can safely only run this check for changed or newly added packs.

packs/witai/actions/text_query.py
```diff
@@ -22,6 +22,7 @@ def run(self, text, application=None):
         if not access_token:
             raise Exception('Missing API key for application %s' % application)
 
+       # pylint: disable=no-member
```
Kami: As the commit message says - sadly `wit` library initializes some of those members lazily based on the underlying C library so pylint is not aware that those members exist.

Sadly not much more we can do here (we could write a custom pylint plugin which tells pylint about all the C library members, but that's obviously a big overkill for now and not something we want to maintain).

.travis.yml
```diff
@@ -6,12 +6,23 @@ os:
 env:
   global:
     - "PIP_DOWNLOAD_CACHE=$HOME/.pip-cache"
+    # Temporary to get PR tests to pass
+    - "ST2_REPO_BRANCH=register_content_fail_on_failure_flag"
```
Kami: This will be removed once that branch is merged into master.

packs/libcloud/actions/start_vm.py
```diff
@@ -18,6 +18,6 @@ def run(self, credentials, vm_id):
         if status is True:
             self.logger.info('Successfully started node "%s"' % (node))
         else:
-            self.logger.rror('Failed to start node "%s"' % (node))
+            self.logger.error('Failed to start node "%s"' % (node))
```
Kami: I fix a bug in the existing pylint script (previously we didn't add the right path to st2 components in `PYTHONPATH`) so we didn't detect any logging related issues.

Kami:
Alright, all the tests finally passed.

I will go ahead, merge this into master and start working on a new PR where I cherry pick "safe" changes from #332 and make relevant improvements and changes there.

PR #343, opened 2015-12-14
Speed up makefile checks

Kami:
This pull request cherry picks changes from #332.

On top of that it includes some fixes and improvements for those changes. Most notably, we now only operate on changed files inside PRs / branchs. When the checks run under master branch we will still run them on all the files. That's a final safe guard in case something goes wrong.

As noted here (https://github.com/StackStorm/st2contrib/pull/332#issuecomment-163088243), using naive file / pack changes approach is not totally safe since other things which can affect on the test / check result can also change (e.g. third party dependency, st2 repo, etc.) and we won't detect those changes.

Hopefully though still running checks on all the files on the master branch should be sufficient to catch those issues while still preserving the original benefit of those changes - providing a better (and faster) experience for the user who is submitting a PR.

### DIFF
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -1,8 +1,16 @@
 #!/usr/bin/env bash
 
+GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
 if [ -z ${TASK} ]; then
   echo "No task provided"
   exit 2
+fi
+
+# When running on master branch we want to run checks on all the files / packs
+# not only on changes ones.
+if [ "${GIT_BRANCH}" = "master" ]; then
+    export FORCE_CHECK_ALL_FILES=true
 fi
 
 if [ ${TASK} == "flake8" ]; then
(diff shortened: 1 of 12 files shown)


scripts/travis.sh
```diff
@@ -1,10 +1,17 @@
...
 if [ -z ${TASK} ]; then
   echo "No task provided"
   exit 2
-  fi
+fi
 
+# When running on master branch we want to run checks on all the files / packs
+# not only on changes ones.
+if [ "${GIT_BRANCH}" = "master" ]; then
+    export FORCE_CHECK_ALL_FILES=true
```
Kami: I will also document in readme how to use this option.

jfryman:
:sunglasses:

Kami:
Alright, pushed more polish and improvements. Going to merge it.

If we get too many false negatives in the future (passes in PR, but fails in master), we can always revert those changes.

Kami:
Thinking about it some more - I will look into creating "daily" builds of master with `FORCE_CHECK_ALL_FILES=true`

Looks like Travis doesn't natively support it, but there are some workarounds for it - https://github.com/travis-ci/travis-ci/issues/582

PR #344, opened 2015-12-14
Reamaze conversations

jfryman:
This PR adds the ability to search through re:amaze conversations in an action.

### DIFF
--- a/packs/reamaze/actions/get_conversations.py
+++ b/packs/reamaze/actions/get_conversations.py
@@ -2,7 +2,22 @@ from lib.actions import BaseAction
 
 
 class GetConversations(BaseAction):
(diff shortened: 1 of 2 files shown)


Kami:
:+1:

Kami:
Hm, checks don't seem OK - need to dig in why git-changes stuff didn't work correctly (it didn't detect any changes, but it works fine locally).

Kami:
I think I found the issue. It turns out `git-changes-*` scripts never worked on Travis.

Travis performs a shallow clone of a single branch which means we can't directly compare to master because master branch / ref is not available locally. I believe this should fix it - https://github.com/StackStorm/st2contrib/commit/5bd2ded50db004a9b79bf6b50e53632deada73ef

jfryman:
@Kami good deal. Thanks for the fix! :heart:

PR #345, opened 2015-12-15
fixed typo in create_vm. 

tonybaloney:
it never would have worked, i tried. for 2 hours...

### DIFF
--- a/packs/libcloud/actions/create_vm.py
+++ b/packs/libcloud/actions/create_vm.py
@@ -1,5 +1,4 @@
 from libcloud.compute.base import NodeSize
-from libcloud.compute.base import NodeImage
 from libcloud.compute.base import NodeLocation
 
 from lib.actions import BaseAction
--- a/packs/libcloud/actions/create_vm.py
+++ b/packs/libcloud/actions/create_vm.py
@@ -46,4 +46,4 @@ class CreateVMAction(BaseAction):
         node = driver.create_node(**kwargs)
 
         self.logger.info('Node successfully created: %s' % (node))
-        return node
+        return self.resultsets.formatter(node)
--- a/packs/libcloud/actions/create_vm.py
+++ b/packs/libcloud/actions/create_vm.py
@@ -32,11 +32,10 @@ class CreateVMAction(BaseAction):
             size = [s for s in driver.list_sizes() if size_name in s.name][0]
 
         if image_id is not None:
-            image = NodeImage(id=image_id, name=None,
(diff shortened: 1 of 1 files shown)


packs/libcloud/actions/create_vm.py
```diff
@@ -32,11 +31,10 @@ def run(self, credentials, name, size_id=None, image_id=None, size_name=None, im
             size = [s for s in driver.list_sizes() if size_name in s.name][0]
 
         if image_id is not None:
-            image = NodeImage(id=image_id, name=None,
-                              driver=driver)
+            image = [i for i in driver.list_images() if image_id == i.id][0]
```
Kami: In theory, existing code should work since the drivers should really only just need an `id` for create_node.

It seems like though that some driver is using some other attribute (name?) and not id.

Kami: Maybe it would also be good to add some logging here so you don't need to waste so much time to figure out the problem.

packs/libcloud/actions/create_vm.py
```diff
@@ -32,11 +31,10 @@ def run(self, credentials, name, size_id=None, image_id=None, size_name=None, im
...
         elif image_name is not None:
             image = [i for i in driver.list_images() if
-                     image_name in i['extra'].get('displaytext', image.name)][0]
```
Kami: Good catch - now that we have testing infrastructure in place we should add some tests for this :)

packs/libcloud/actions/create_vm.py
tonybaloney: some of the drivers, assume that the image was fetched using list_images or get_image. and look for extra[stuff] in order to make the request. GCE is a prime example

Kami: Ah, good point.

And yeah, using `list_images` tends to be safer, but in some cases it adds a lot of overhead (list_image responses can be massive) so if you already know the id and want to avoid additional request that's the preferred approach.

Kami: Also, now that I think of it, you probably got `AttributeError` / `KeyError` when trying to create a GCE node so I see how debugging this could waste a bunch of time :/

packs/libcloud/actions/create_vm.py
tonybaloney: yeah i was close to running pdb on the actionrunner. then i spotted it

Kami:
Merged, thanks.

PR #346, opened 2015-12-15
Feature/create node kwargs

tonybaloney:
When users are running create_vm it's useful to define defaults for some drivers. this would be in create_node_kwargs within the config.yaml for each driver setting.

### DIFF
--- a/packs/libcloud/actions/create_vm.py
+++ b/packs/libcloud/actions/create_vm.py
@@ -1,5 +1,4 @@
 from libcloud.compute.base import NodeSize
-from libcloud.compute.base import NodeImage
 from libcloud.compute.base import NodeLocation
 
 from lib.actions import BaseAction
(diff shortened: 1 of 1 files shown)


Kami:
I would personally prefer for those (additional create_node kwargs) to be an action parameter.

This way it's easier to re-use and test it (action params are almost always preferred since they make code more re-usable and easier to test).

tonybaloney:
both would be nice too.... having defaults removes the need to constantly parse them around.

Kami:
If you want to bind action parameter to some particular value, one thing you can do is create a wrapper action metadata file which binds that parameter to a value. And if you want you can also make that parameter immutable.

For example, something like:

```yaml
---
name: gce_create_vm
runner_type: run-python
description: Create a new VM in GCE.
enabled: true
entry_point: create_vm.py
parameters:
  credentials:
    type: string
    description: Name of the credentials set (as defined in the config) to use.
    immutable: true
    default: "gce"
....
  extra_kwargs:
    type: object
    description: ....
    required: false
    default: { "ex_foo": "blah" }
...
```

PR #349, opened 2015-12-16
Add new time pack

Kami:
It includes initial `get_week_boundaries` action for retrieving week boundaries.

It's a functionality I need for the github activity email, but I think it's something which may come useful in other scenarios as well so adding it to a general purpose pack makes some sense.

TODO:

- [x] Tests
- [x] Icon


packs/time/pack.yaml
```diff
@@ -0,0 +1,9 @@
+---
+name: time
```
Kami: I'm also open for a better name. date? date_time? dunno :P

jfryman: chronos?

Kami:
We had a quick discussion on Slack if this warrants a pack or not.

I think it makes sense since it might come handy in other scenarios as well. On top of that, I could see other actions such as "parse_date" which parses a human readable date string and returns a timestamp (or similar) also live in this pack.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450289918014987

PR #351, opened 2015-12-18
Update github.list_issues

Kami:
Make sure author, assignee and closed_by issues are dict so we can include more info / attributes about those user.


Kami:
@emedvedev This might affect you.

PR #353, opened 2015-12-24
process channel_joined message subtype

jfryman:
This PR updates the `_process_message` function to have some logic to process a message subtype (https://api.slack.com/events/message) so they are not discarded by the mainline logic.

### DIFF
--- a/packs/slack/sensors/slack_sensor.py
+++ b/packs/slack/sensors/slack_sensor.py
@@ -143,7 +143,8 @@ class SlackSensor(PollingSensor):
             # Skip unsupported event
             return
 
-        if 'subtype' in data and data['subtype'] in self._message_subtype_handlers:
+        subtype = data.get('subtype', None)
+        if subtype in self._message_subtype_handlers:
             subtype = data['subtype']
             handler = self._message_subtype_handlers.get(subtype)
             handler(data)
@@ -168,29 +169,17 @@ class SlackSensor(PollingSensor):
           else:
               text = data['text']
 
-          payload = {
-              'user': {
-                  'id': user_info['id'],
-                  'name': user_info['name'],
-                  'first_name': user_info['profile'].get('first_name',
-                                                        'Unknown'),
-                  'last_name': user_info['profile'].get('last_name',
-                                                        'Unknown'),
-                  'real_name': user_info['profile'].get('real_name',
-                                                        'Unknown'),
-                  'is_admin': user_info['is_admin'],
-                  'is_owner': user_info['is_owner']
-              },
+          common_payload = self._get_message_common_payload(data)
+          channel_payload = {
               'channel': {
                   'id': channel_info['id'],
                   'name': channel_info['name'],
                   'topic': channel_info['topic']['value'],
                   'is_group': channel_info.get('is_group', False),
               },
-              'timestamp': int(float(data['ts'])),
-              'timestamp_raw': data['ts'],
-              'text': text
           }
+          payload = common_payload.copy()
+          payload.update(channel_payload)
 
           self._sensor_service.dispatch(trigger=trigger, payload=payload)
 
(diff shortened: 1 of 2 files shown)


packs/slack/sensors/slack_sensor.yaml
```diff
@@ -20,3 +20,17 @@
             type: "integer"
           timestamp_raw:
             type: "string"
+    -
+      name: "message_channel_join"
```
Kami: Maybe just `join` or perhaps `channel.join` (I find the message part confusing)?

Kami:
This sensor seems like a good candidate for some unit tests with fixtures, heh :)

packs/slack/sensors/slack_sensor.py
```diff
@@ -140,51 +143,56 @@ def _handle_message(self, data):
...
-            channel_info = self._get_group_info(group_id=channel_id)
-
-        if not user_info or not channel_info:
-            # Deleted user or channel
-            return
-
-        # Removes formatting from messages if enabled by the user in config
-        if self._strip_formatting:
-            text = re.sub("<http.*[|](.*)>", "\\1", data['text'])
+        if 'subtype' in data and data['subtype'] in self._message_subtype_handlers:
```
Kami: Minor style thing - `subtype - data.get('subtype', None'); if subtype in self._message...`

jfryman:
> This sensor seems like a good candidate for some unit tests with response fixtures, heh :)

Nice. Let me take a pass.

packs/slack/sensors/slack_sensor.yaml
jfryman: My thinking is that it's a subtype. What I really wanted to do was namespace it as such (`message.channel_join`), but I really was unsure what the effects would be. So, just went with the underscores.

Think I could get away with the namespace?

packs/slack/sensors/slack_sensor.py
```diff
@@ -237,3 +245,27 @@ def _api_call(self, method, **kwargs):
         result = self._client.api_call(method, **kwargs)
         result = json.loads(result)
         return result
+
+    def _handle_message_channel_join(self, data):
+        trigger = 'slack.message_channel_join'
+        user_info = self._get_user_info(user_id=data['user'])
+
+        payload = {
```
Kami: Looks like we could refactor this into `_get_common_payload` method or similar.

packs/slack/sensors/slack_sensor.py
```diff
@@ -237,3 +245,27 @@ def _api_call(self, method, **kwargs):
...
+                'last_name': user_info['profile'].get('last_name',
+                                                      'Unknown'),
+                'real_name': user_info['profile'].get('real_name',
+                                                      'Unknown'),
+                'is_admin': user_info['is_admin'],
+                'is_owner': user_info['is_owner']
+            },
+            'timestamp': int(float(data['ts'])),
+            'timestamp_raw': data['ts'],
+            'text': text
```
Kami: There should be no `text` for this event, right?

packs/slack/sensors/slack_sensor.py
jfryman: subtract (`-`) or assign (`=`)? It's good practice to have an assignment + control on the same line like that?

packs/slack/sensors/slack_sensor.py
jfryman: :+1:

packs/slack/sensors/slack_sensor.py
Kami: Sorry, meant to write `=`. And it's better and nice to split it across two lines, I just did it in a single line because I was too lazy to use code block syntax :)

packs/slack/sensors/slack_sensor.py
jfryman: No, there will still be text for this. We don't actually get `channel` for this event... only a text with `<UID|bob> has joined #channel`. 

WCYD?

PR #354, opened 2015-12-30
Update SmartThings Payload to use API Key

jfryman:
Updates to appropriate HTTP header for API authentication per https://docs.stackstorm.com/authentication.html?highlight=apikey#api-key-usage


Kami:
Good catch - I actually also thought myself that we use `X-` prefix, but it turns it we don't (IIRC, manas linked to an article on why he doesn't like X- prefix, but I forgot we decided to use st2-) :)

manasdk:
https://tools.ietf.org/html/rfc6648

PR #355, opened 2016-01-05
added project parameter to create_issue action

psychopenguin:
This allows one to override the default project from config file, now, anyone can use something like this:

`st2 action execute jira.create_issue description="testing st2" summary="stackstorm test ticket" project="TESTPROJECT"`

### DIFF
--- a/packs/jira/actions/create_issue.py
+++ b/packs/jira/actions/create_issue.py
@@ -7,8 +7,11 @@ __all__ = [
 
 
 class CreateJiraIssueAction(BaseJiraAction):
-    def run(self, summary, type, description=None):
-        project = self.config['project']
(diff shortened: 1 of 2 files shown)


packs/jira/actions/create_issue.py
```diff
@@ -7,8 +7,11 @@
 
 
 class CreateJiraIssueAction(BaseJiraAction):
-    def run(self, summary, type, description=None):
-        project = self.config['project']
+    def run(self, summary, type, description=None, project=None):
+        if project:
+            project = project
```
manasdk: project = project or self.config['project']

I guess that is more pythonic or some such thing.

manasdk:
LGTM. Will merge once suggestion is picked up and CI passes.

PR #361, opened 2016-01-08
Allow for use with elasticsearch 2.1.x.

pixelrebel:
Tested actions:
```
indices.show
indices.snapshot
snapshots.delete
snapshots.show
search.body
search.q
```
All verified to work.  Should we test some more actions, or just grab our nuts and go?


lakshmi-kannan:
I think we should bump the minor version here https://github.com/StackStorm/st2contrib/blob/master/packs/elasticsearch/pack.yaml#L8

So it should be 0.2.0

lakshmi-kannan:
And a note in README.md in the pack describing this change and what to expect would be super helpful for other people. cc: @pixelrebel

PR #362, opened 2016-01-13
Getting rid of immutable flag for sudo parameters in ansible pack

emptywee:
- bumping version to 0.3
 - removing immutable flag for 'sudo' param of all actions


armab:
:+1: 

Also small note in https://github.com/emptywee/st2contrib/blob/master/packs/ansible/CHANGES.md would be good.

Kami:
LGTM, thanks.

PR #365, opened 2016-01-15
Basic updates to make handler work with https

manasdk:


Kami:
IIRC, we have a similar script lying around somewhere?

Or, maybe I'm mixing things up.

Kami:
Now that we have a test framework in place we should also add some tests for this script since it always gets out of date or it's broken.

manasdk:
> IIRC, we have a similar script lying around somewhere?

Isn't this the script?

manasdk:
> Now that we have a test framework in place we should also add some tests for this script since it always gets out of date or it's broken.

Needs integration tested against a full stackstorm install. Is that possible or only unit testing?

DoriftoShoes:
So this looks good but when did we make the single port change?  Are the old URLs still in any currently supported versions?

manasdk:
The single port change went in v1.2. I don't thing old URLs are supported any longer

manasdk:
> I don't thing old URLs are supported any longer

This is an installer choice though so you need depends on the configuration.

estee-tew:
https://stackstorm.slack.com/archives/support/p1454967684000368

PR #305, opened 2015-10-28
Add support for additional AWS services

DoriftoShoes:
This PR adds support for the following AWS services:

- [x] IAM
- [x] VPC
- [x] S3
- [x] CloudFormation
- [x] RDS
- [x] SQS

Known issues:

- [x] parsers need to be fixed

PR #366, opened 2016-01-18
Import new SQS sensor for AWS integration pack

haad:
Add new sensor for incoming SQS messages and add autogenerated SQS actions.

### DIFF
--- a/packs/aws/requirements.txt
+++ b/packs/aws/requirements.txt
@@ -1,1 +1,2 @@
 boto
+boto3
--- a/packs/aws/sensors/sqs_sensor.py
+++ b/packs/aws/sensors/sqs_sensor.py
@@ -1,6 +1,6 @@
 """
 This is generic SQS Sensor using boto3 api to fetch messages from sqs queue.
-After receiving a messge it's content is passed as payload to a trigger 'aws.sqs_new_message'
+After receiving a message it's content is passed as payload to a trigger 'aws.sqs_new_message'
 
 This sensor can be configured either by using config.yaml withing a pack or by creating
 following values in datastore:
@@ -8,6 +8,16 @@ following values in datastore:
     - aws.aws_access_key_id
     - aws.aws_secret_access_key
     - aws.region
+
+For configuration in config.yaml with config like this
+    setup:
+      aws_access_key_id:
+      aws_access_key_id:
+      region:
+    sqs_sensor:
+      input_queue:
+
+If any value exist in datastore it will be taken instead of any value in config.yaml
 """
 
 from boto3.session import Session
@@ -22,18 +32,18 @@ class AWSSQSSensor(PollingSensor):
                                            poll_interval=poll_interval)
 
     def setup(self):
-        self.input_queue = self._GetConfigEntry(key='input_queue', prefix='sqs_sensor')
-        self.aws_access_key = self._GetConfigEntry('aws_access_key_id')
-        self.aws_secret_key = self._GetConfigEntry('aws_secret_access_key')
-        self.aws_region = self._GetConfigEntry('region')
+        self.input_queue = self._get_config_entry(key='keys ', prefix='sqs_sensor')
+        self.aws_access_key = self._get_config_entry('aws_access_key_id')
+        self.aws_secret_key = self._get_config_entry('aws_access_key_id')
+        self.aws_region = self._get_config_entry('region')
 
         self._logger = self._sensor_service.get_logger(name=self.__class__.__name__)
 
         self.session = None
         self.sqs_res = None
 
-        self._SetupSqs()
-        self.queue = self._GetQueueByName(self.input_queue)
+        self._setup_sqs()
+        self.queue = self._get_queue_by_name(self.input_queue)
 
     def poll(self):
         msg = self._receive_messages(queue=self.queue)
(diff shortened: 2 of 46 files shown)


Kami:
Thanks for the contribution, we will look into it shortly.

Kami:
I assume the actions were automatically generated, right?

packs/aws/requirements.txt
```diff
@@ -1 +1,2 @@
 boto
+boto3
```
Kami: Just curious, why do we also need `boto3`?

It looks like sensor uses it. I would assume SQS is not supported in `boto`, or something else?

packs/aws/sensors/sqs_sensor.py
```diff
@@ -0,0 +1,95 @@
+"""
+This is generic SQS Sensor using boto3 api to fetch messages from sqs queue.
+After receiving a messge it's content is passed as payload to a trigger 'aws.sqs_new_message'
+
+This sensor can be configured either by using config.yaml withing a pack or by creating
+following values in datastore:
+    - aws.input_queue
```
Kami: For consistency with other sensor, it would be good to nest those options under `sqs_sensor`.

So something like that:

```yaml
service_notifications_sensor:
  host: "localhost"
  port: 12345
  path: "/my-path"

sqs_sensor:
  input_queue: "fooo"
```

packs/aws/sensors/sqs_sensor.py
```diff
@@ -0,0 +1,95 @@
...
+    - aws.aws_access_key_id
```
Kami: It looks like those attributes (access key id, access key secret) and region should already be available under `setup` key - that's what actions use.

See `config.txt` for details.

packs/aws/sensors/sqs_sensor.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self._SetupSqs()
+        self.queue = self._GetQueueByName(self.input_queue)
+
+    def poll(self):
+        msg = self._receive_messages(queue=self.queue)
+        if msg:
+            payload = {"queue": self.input_queue, "body": msg[0].body}
+            self._sensor_service.dispatch(trigger="aws.sqs_new_message", payload=payload)
+            msg[0].delete()
```
Kami: Probably a good idea to wrap delete in try / except or is exception bubbling up desired?

Also, I assume "delete" is basically ACK?

packs/aws/sensors/sqs_sensor.py
```diff
@@ -0,0 +1,95 @@
...
+        pass
+
+    def update_trigger(self, trigger):
+        # This method is called when trigger is updated
+        pass
+
+    def remove_trigger(self, trigger):
+        pass
+
+    def _GetConfigEntry(self, key, prefix='setup'):
```
Kami: Minor style thing - please use underscore separated method names - `_get_config_entry`, `_setup_sqs`, etc.

Kami:
I've added some comments.

Thanks again!

manasdk:
Some SQS stuff also in https://github.com/StackStorm/st2contrib/pull/305. Btw, it is that PRs problem to deal with merge conflicts since @DoriftoShoes has yet to take away the WIP label :)

packs/aws/requirements.txt
haad: I already had this code written for boto3 rewriting it for boto would be more time consuming.

packs/aws/sensors/sqs_sensor.py
haad: That's how is it done now, maybe I have forgot to push it let me check

packs/aws/sensors/sqs_sensor.py
haad: Yes they can be there.

packs/aws/sensors/sqs_sensor.py
haad: Ok will do.

packs/aws/sensors/sqs_sensor.py
haad: ok will do

packs/aws/requirements.txt
Kami: That's fine, lets just add a comment to the requirements.txt file which says that boto3 is used by SQS sensor.

packs/aws/sensors/sqs_sensor.py
Kami: Ah, OK, I think I missed the datastore part.

It would also be good to explicitly document which values have precedence (config or datastore).

packs/aws/sensors/sqs_sensor.py
haad: It will just let SQS know that it this message was processed and it can remove that message from queue. According to [http://boto3.readthedocs.org/en/latest/reference/services/sqs.html#SQS.Message.delete | doc] I'm not sure if it will raise anything at all.

Kami: Alright, so yeah - it behaves like ACK.

In this case the current approach is fine - we only want to ACK if the message has already been processed. If the processing fails (exception is thrown or similar) message won't be ACKed so it will be re-delivered (re-retrieved) and processed again.

haad:
ok so I think I have fixed everything except handling delete exceptions but I have not find any mention about raising anything inside of delete routine so I'm not sure there.

Kami:
Looks good.

I will go ahead and merge it - thanks!

PR #367, opened 2016-01-25
St2 pack requirements

manasdk:


Kami:
LGTM, +1

enykeev:
Still think there's a value in making it a core pack so it would share the venv with the rest of st2.

Kami:
I think real long term solution is implementing pack versioning, but making it a core pack would also work in the short term.

PR #371, opened 2016-01-27
Add multiple queues support to AWS sqs sensor

haad:
Add support for multiple queues into sqs sensor to make sure we can work with multiple rules. Raise an error if we don't get any configuration from ds nor config.yaml

### DIFF
--- a/packs/aws/sensors/sqs_sensor.py
+++ b/packs/aws/sensors/sqs_sensor.py
@@ -4,7 +4,7 @@ After receiving a message it's content is passed as payload to a trigger 'aws.sq
 
 This sensor can be configured either by using config.yaml withing a pack or by creating
 following values in datastore:
-    - aws.input_queue
+    - aws.input_queues (list queues as comma separrated string: first_queue,second_queue)
     - aws.aws_access_key_id
     - aws.aws_secret_access_key
     - aws.region
@@ -15,7 +15,9 @@ For configuration in config.yaml with config like this
       aws_access_key_id:
       region:
     sqs_sensor:
-      input_queue:
+      input_queues:
+        - first_queue
+        - second_queue
 
 If any value exist in datastore it will be taken instead of any value in config.yaml
 """
@@ -32,9 +34,17 @@ class AWSSQSSensor(PollingSensor):
                                            poll_interval=poll_interval)
 
     def setup(self):
-        self.input_queue = self._get_config_entry(key='keys ', prefix='sqs_sensor')
+        queues = self._get_config_entry(key='input_queues', prefix='sqs_sensor')
+
+        # XXX: This is a hack as from datastore we can only receive a string while
+        # from config.yaml we can receive a list
+        if type(queues) is str:
+            self.input_queues = queues.split(', ')
+        else:
+            self.input_queues = queues
+
         self.aws_access_key = self._get_config_entry('aws_access_key_id')
-        self.aws_secret_key = self._get_config_entry('aws_access_key_id')
+        self.aws_secret_key = self._get_config_entry('aws_secret_access_key')
         self.aws_region = self._get_config_entry('region')
 
         self._logger = self._sensor_service.get_logger(name=self.__class__.__name__)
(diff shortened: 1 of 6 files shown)


packs/aws/sensors/sqs_sensor.py
```diff
@@ -32,9 +34,17 @@ def __init__(self, sensor_service, config=None, poll_interval=5):
                                            poll_interval=poll_interval)
 
     def setup(self):
-        self.input_queue = self._get_config_entry(key='keys ', prefix='sqs_sensor')
+        queues = self._get_config_entry(key='input_queues', prefix='sqs_sensor')
```
lakshmi-kannan: Can you please handle 'input_queue' as well? Otherwise, this might break existing users. Add a LOG.warning() to indicate that the config is deprecated. We can remove it in next version.

packs/aws/sensors/sqs_sensor.py
```diff
@@ -32,9 +34,17 @@ def __init__(self, sensor_service, config=None, poll_interval=5):
...
+
+        # XXX: This is a hack as from datastore we can only receive a string while
+        # from config.yaml we can receive a list
+        if type(queues) is str:
```
m4dcoder: I recommend changing this check to the following.

import six
...
isinstance(queues, six.string_types)

lakshmi-kannan:
Also, pack.yaml version needs to be bumped. A CHANGELOG in README.md would be nice for people to get a gist of changes.

packs/aws/sensors/sqs_sensor.py
```diff
@@ -32,9 +34,17 @@ def __init__(self, sensor_service, config=None, poll_interval=5):
...
+            self.input_queues = queues.split(', ')
```
m4dcoder: I recommend covering cases where there are no whitespace or multiple whitespaces before/after the comma.

packs/aws/sensors/sqs_sensor.py
```diff
@@ -32,9 +34,17 @@ def __init__(self, sensor_service, config=None, poll_interval=5):
...
+        else:
```
m4dcoder: If it's not a string, I suggest raising a ValueError that it's not in the expected format.

haad: As I'm saying in comment problem is that from DS I receive a string and from config.yaml an array so both are valid initially.

haad:
Ok so I will bump version in pack.yaml to 0.4 or to 0.3.1 ?

packs/aws/sensors/sqs_sensor.py
haad: This sensor is in repo only for 2-3 days. I did initial PR too. I'm not sure if anyone else it used yet. So is this really necessary ?

lakshmi-kannan:
@haad Bump to 0.4.0 please

packs/aws/sensors/sqs_sensor.py
lakshmi-kannan: I was recommending the ideal approach. We can skip this since it's only 3 days.

packs/aws/sensors/sqs_sensor.py
lakshmi-kannan: Good point.http://stackoverflow.com/questions/4071396/split-by-comma-and-strip-whitespace-in-python

packs/aws/sensors/sqs_sensor.py
haad: Thanks :)

packs/aws/sensors/sqs_sensor.py
```diff
@@ -32,9 +35,17 @@ def __init__(self, sensor_service, config=None, poll_interval=5):
...
 
     def setup(self):
-        self.input_queue = self._get_config_entry(key='keys ', prefix='sqs_sensor')
+        queues = self._get_config_entry(key='input_queues', prefix='sqs_sensor')
+
+        # XXX: This is a hack as from datastore we can only receive a string while
+        # from config.yaml we can receive a list
+        if isinstance(queues, six.string_types):
+            self.input_queues = [x.strip() for x in queues.split(',')]
+        else:
```
m4dcoder: What are the use cases when queues is not a string type?  Otherwise, why can't we raise an Exception here instead of continue processing?

lakshmi-kannan: input_queues can be a YAML list instead of a comma separated string. Sure someone can specify an object or boolean or something crazy but I think this code is fine.

lakshmi-kannan: BTW https://github.com/haad/st2contrib/blob/aws_sqs_sensor/packs/aws/requirements.txt is missing six as a dep.

PR #373, opened 2016-01-28
Update get_alerts.py

lanthos:
Current super was causing a TypeError:

super(GetAppHealthStatusAction, Action).__init__(*args, **kwargs)
TypeError: super(type, obj): obj must be an instance or subtype of type

Changing to super(GetAppHealthStatusAction, self) resolves the issue and things are working correctly again.


Kami:
Good catch & thanks!

PR #374, opened 2016-01-29
Feature/GitHub users

tonybaloney:
Add new actions 
* Get user
* list teams for organization
* Add (and invite if not member) user to team within organization

Bug exists in the current version of PyGitHub, so using their trunk.

### DIFF
--- a/packs/github/actions/add_team_membership.yaml
+++ b/packs/github/actions/add_team_membership.yaml
@@ -9,6 +9,10 @@ parameters:
     type: "string"
     description: "User to invite"
     required: true
+  organization:
+    type: "string"
+    description: "Organization the team belongs to"
+    required: true
   team:
     type: "string"
     description: "Team ID."
--- a/packs/github/actions/add_team_membership.yaml
+++ b/packs/github/actions/add_team_membership.yaml
@@ -13,7 +13,7 @@ parameters:
     type: "string"
     description: "Organization the team belongs to"
     required: true
-  team:
+  team_id:
     type: "integer"
     description: "Team ID."
     required: true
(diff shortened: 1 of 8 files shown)


Kami:
Great, thanks.

And yeah, last time I checked, PyGithub was less than ideal. In fact, all of the Python libraries for interacting with Github I checked were less than ideal (kinda sad that Github doesn't develop an official one, anyways).

packs/github/actions/add_team_membership.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: add_team_membership
```
Kami: add_team_member perhaps?

Anyway, not a big deal, I'm also fine with the current name :)

packs/github/actions/add_team_membership.yaml
```diff
@@ -0,0 +1,19 @@
...
+parameters:
+  user:
+    type: "string"
+    description: "User to invite"
+    required: true
+  organization:
+    type: "string"
+    description: "Organization the team belongs to"
+    required: true
+  team:
```
Kami: Probably better to be explicit and call it `team_id`, but would need to check existing actions. Maybe they don't use `id` suffix and current approach is better for consistency,

packs/github/actions/add_team_membership.yaml
tonybaloney: add_team_member is an existing action, which is being deprecated in the GitHub API. Depends on whether the user knows the github API!

manasdk:
+1

Merging.

manasdk:
@tonybaloney Appreciate the contribution!

PR #375, opened 2016-01-29
Catching when first_name is None, since it is optional 

tonybaloney:
(should it be optional)?


emedvedev:
I'd not include `first_name` in the request at all if it's `None`, but that works, too. :) Looks good.
Could you please bump the pack version, too?

emedvedev:
Thanks. :) Merging.

PR #376, opened 2016-01-29
Allow user to specify "cacert" config option in the st2 pack

Kami:
By default, if `cacert` option is not passed to the client constructor (current behavior), the client doesn't perform cert validation (https://github.com/StackStorm/st2/blob/master/st2client/st2client/utils/httpclient.py#L30).

If a user is using a cert issued by trusted CA authority or want to turn cert validation on they can do this by specifying `cacert` config option (either path to CA bundle or True to use CA from system store).


Kami:
I merged the corresponding st2 PR, but I need to wait with this one until a new version of StackStorm and st2client is released and published.

Kami:
This feature has been released in StackStorm v1.3.1 so I'm merging this into master.

PR #377, opened 2016-01-29
[WIP] Adding auth_url attribute

jfryman:
This PR adds a new `auth_url` attribute to the `st2` base pack. This allows an overwrite to ensure that the both authentication and authorization URLs can be specified independently.

This is to accommodate the move of the endpoints to subdirectories in the AIO versus individual ports.

Working with Justin in #stackstorm-community.

### DIFF
--- a/packs/st2/actions/lib/action.py
+++ b/packs/st2/actions/lib/action.py
@@ -28,8 +28,8 @@ class St2BaseAction(Action):
     def _get_st2_urls(self):
         # First try to use base_url from config.
         base_url = self.config.get('base_url', None)
-        api_url = None
-        auth_url = None
+        api_url = self.config.get('api_url', None)
+        auth_url = self.config.get('auth_url', None)
 
         # not found look up from env vars. Assuming the pack is
         # configuered to work with current StackStorm instance.
--- a/packs/st2/actions/lib/action.py
+++ b/packs/st2/actions/lib/action.py
@@ -20,21 +20,24 @@ class St2BaseAction(Action):
         self.client = self._get_client()
 
     def _get_client(self):
-        base_url, api_url = self._get_st2_urls()
+        base_url, api_url, auth_url = self._get_st2_urls()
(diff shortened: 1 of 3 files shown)


jfryman:
Tested working.

Kami:
Probably going to conflict with https://github.com/StackStorm/st2contrib/pull/376, but not a big deal, easy enough to resolve conflicts :)

packs/st2/actions/lib/action.py
```diff
@@ -20,21 +20,24 @@ def __init__(self, config):
...
         base_url = self.config.get('base_url', None)
-        api_url = None
+        api_url = self.config.get('api_url', None)
+        auth_url = self.config.get('auth_url', None)
 
         # not found look up from env vars. Assuming the pack is
         # configuered to work with current StackStorm instance.
         if not base_url:
             api_url = os.environ.get('ST2_ACTION_API_URL', None)
+            auth_url = os.environ.get('ST2_ACTION_AUTH_URL', None)
```
Kami: I just noticed we don't actually set this variable in st2 for actions.

Not a big deal, but something to look into.

packs/st2/actions/lib/action.py
```diff
@@ -20,21 +20,24 @@ def __init__(self, config):
...
-        base_url, api_url = self._get_st2_urls()
+        base_url, api_url, auth_url = self._get_st2_urls()
         token = self._get_auth_token()
-        return self._client(base_url=base_url, api_url=api_url, token=token)
+        return self._client(base_url=base_url, api_url=api_url,
+                            auth_url=auth_url, token=token)
 
     def _get_st2_urls(self):
         # First try to use base_url from config.
         base_url = self.config.get('base_url', None)
```
Kami: Also, I'm a bit confused.

Did you intentionally remove `base_url` from the config? Because we still retrieve it from the config here so it looks like we still support that option.

PR #386, opened 2016-02-05
Fix aws result formatter

m4dcoder:


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454640331005179

packs/aws/actions/lib/ec2parsers.py
```diff
@@ -35,6 +130,8 @@ def selector(self, output):
             return self.parseInstance(output)
         elif isinstance(output, boto.ec2.volume.Volume):
             return self.parseVolume(output)
+        elif isinstance(output, boto.ec2.blockdevicemapping.BlockDeviceType):
```
manasdk: since BlockDeviceType is an object(https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/ec2/blockdevicemapping.py#L25) can we use some generic mechanism to parse all python objects.

Writing out FieldList like https://github.com/StackStorm/st2contrib/pull/386/files#diff-417c109ec13ece08b2872bd616c2e1f3R17 leads to too much special casing unfortunately.

m4dcoder: Let's move the parser rework as a separate PR for now so we don't block here.

manasdk: Created issue https://github.com/StackStorm/st2contrib/issues/387 for tracking. Since marshalling and unmarshalling object properly is my pet-peeve I might take a crack at it :P

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1455084860006236

PR #393, opened 2016-02-10
AWS pack formatter fixes

Kami:
This pull request adds back changes from #386 which were previously reverted to get things working.

In addition to that it also adds changes file where this backward incompatible change is described. I will also add a knowledge base entry which points to this change file. We should link users to that entry in case this encounter this issue after this change has been merged.

Note: We should only merge this after v1.3.1 is out.

### DIFF
--- a/packs/aws/CHANGES.rst
+++ b/packs/aws/CHANGES.rst
@@ -1,0 +1,10 @@
+## v0.6.0
+
+* Fix the result format of some of the actions such as ``ec2_run_instances``. Previously,
+  the result was a list of lists of dicts and now the result is correctly a list of dicts.
+
+  Keep in mind that this is a breaking changes.
+
+  If you previously accessed the result of the ``ec2_run_instances`` action in the action-chain
+  workflow like that - ``run_instances.result[0][0].id``, you need to update it so it looks like
+  this ``run_instance.result[0].id``.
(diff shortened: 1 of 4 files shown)


Kami:
I've created a KB article for those changes - http://stackstorm.reamaze.com/kb/packs/dict-object-has-no-element-0-or-similar-error-when-using-an-action-from-aws-pack

If people encounter issues after they upgrade / reinstall aws pack, please point them to that article. Also feel free to improve the wording and add more info to it.

Kami:
Looks like there are more issues related to those changes, looking into it.

```python
  stderr: "Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/st2actions/runners/python_action_wrapper.py", line 116, in <module>
    obj.run()
  File "/usr/lib/python2.7/dist-packages/st2actions/runners/python_action_wrapper.py", line 61, in run
    output = action.run(**self._parameters)
  File "/root/st2contrib/packs/aws/actions/run.py", line 18, in run
    return self.do_method(module_path, cls, action, **kwargs)
  File "/root/st2contrib/packs/aws/actions/lib/action.py", line 94, in do_method
    resultset = getattr(obj, action)(**kwargs)
AttributeError: 'NoneType' object has no attribute 'run_instances'
```

PR #395, opened 2016-02-12
initial kubernetes pack with thirdpartyresource integration

mward29:
This is an initial Kubernetes integration for ThirdPartyResource. It can watch the Kubernetes API and and trigger and AWS RDS database creation or deletion based.


packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,93 @@
...
+        name = d_list['object']['metadata']['name']
+        namespace = d_list['object']['metadata']['namespace']
+        uid = d_list['object']['metadata']['uid']
+        # Now lets see if labels exist, if so build a trigger
+        if 'labels' in d_list['object']['metadata']:
+            labels_data = d_list['object']['metadata']['labels']
+            self._build_a_trigger(resource_type=resource_type, name=name, labels=labels_data,
+                                  namespace=namespace, object_kind=object_kind, uid=uid)
+        else:
+            print("No Labels for the resource below. Tough to proceed without knowing how \
```
tonybaloney: def use a logger instead of a print statement, since this won't go

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,93 @@
...
+            'uid': uid
+        }
+        self._logger = self._sensor_service.get_logger(name=self.__class__.__name__)
+        self._logger.debug('Triggering Dispatch Now')
+
+        # Create dispatch trigger
+        self._sensor_service.dispatch(trigger=trigger, payload=payload)
+        pass
+
+    def cleanup(self):
```
tonybaloney: you can just remove those altogether

packs/kubernetes/requirements.txt
```diff
@@ -0,0 +1 @@
+pykube
```
tonybaloney: probably choose a specific version unless you plan to maintain back compat with the pack

packs/kubernetes/actions/db_rds_spec.py
```diff
@@ -0,0 +1,39 @@
...
+        db_name = re.sub('[^0-9a-zA-Z]+', '-', payload['name']) + "-" + payload['namespace']
+
+        # Lets get a username generated
+        user_name = self._user_name(uid=payload['uid'])
+        # Lets get a password randomly generated
+        pw = self._id_generator()
+        instance_class = 'db.m3.xlarge'
+        port = '3306'
+        rds_engine = 'mysql'
+        allocated_storage = '50'
```
tonybaloney: Is this configurable?

packs/kubernetes/actions/db_rds_spec.py
```diff
@@ -0,0 +1,39 @@
+import re
+import uuid
+from st2actions.runners.pythonrunner import Action
```
lakshmi-kannan: newline please to "group" imports :P.

packs/kubernetes/actions/db_rds_spec.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class DatabaseRdsSpec(Action):
+    def run(self, payload):
+        # For dbidentifier
+        db_name = re.sub('[^0-9a-zA-Z]+', '-', payload['name']) + "-" + payload['namespace']
```
lakshmi-kannan: Some comment here with an example would definitely help.

packs/kubernetes/actions/db_rds_spec.py
```diff
@@ -0,0 +1,39 @@
...
+            'rds_engine': rds_engine,
+            'allocated_storage': allocated_storage,
+        }
+
+        return payload
+
+    def _user_name(self, uid):
+        short_uid = uid[0:7]
+        return "db_" + short_uid
+        pass
```
lakshmi-kannan: Remove pass in line 35 and 39.

packs/kubernetes/actions/db_rds_spec.py
lakshmi-kannan: Just a note about python dicts, the [] notation throws a KeyError when key does not exist. For example, payload['name'] would throw. Sometimes you can do payload.get('name', default_name) or wrap the lines of code in try catch block.

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def run(self):
+        self._logger = self._sensor_service.get_logger(name=self.__class__.__name__)
+        self._logger.debug('Watch Kubernetes for thirdpartyresource information')
+        r = self.client
+        lines = r.iter_lines()
+        # Save the first line for later or just skip it
+#        first_line = next(lines)
+        for line in lines:
+            io = json.dumps(line)
```
lakshmi-kannan: Little sketchy since dumps and loads are inverse operations.

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,93 @@
...
+            n = json.loads(io)
+            d_list = ast.literal_eval(n)
+            self._k8s_object(d_list=d_list)
+#            self._resource_version(d_list=d_list)
+        pass
```
lakshmi-kannan: Remove all the `pass` lines at end of functions if they have actual code.

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,93 @@
...
+        self._logger.debug('Watch Kubernetes for thirdpartyresource information')
+        r = self.client
+        lines = r.iter_lines()
+        # Save the first line for later or just skip it
+#        first_line = next(lines)
+        for line in lines:
+            io = json.dumps(line)
+            n = json.loads(io)
+            d_list = ast.literal_eval(n)
+            self._k8s_object(d_list=d_list)
```
lakshmi-kannan: I'd rather make this function return an object and dispatching a trigger should happen here. That way code is better unit testable.

mward29: It was intentional to not place the entirety of _k8s_object function in the for loop. I can certainly change this but I'm partial to segregating bits of code up into understandable chunks. If you have a recommended standard by which to not have one mass of code all in a single for loop I'm all ears.

packs/kubernetes/actions/db_rds_spec.py
mward29: Should be yes. I'm figuring out how to best address this. I'd like to place something like the below in config.yaml. Allowing us to rename db_rds_spec and use it as a more general piece of code for parsing through data sets.

    rds:
      size: "db.m3.xlarge"
      port: "3306"
      object: "mysql"
      storage: "100"

I'd like to use the keys above as a way to search for values in the labels. Something like payload['labels']['size']. If does not exist, use the value above as default.

packs/kubernetes/actions/db_rds_spec.py
mward29: Assuming this would cause the action-chain to exit. That would be the intended outcome. This action-chain should not continue without it.

packs/kubernetes/actions/db_rds_spec.py
tonybaloney: then do
```python
labels = config.get('labels', None)
if labels is not None:
  size = labels.get('size', 'db.m3.large')
else:
 size = 'db.m3.large'
```
Or something like that

packs/kubernetes/sensors/thirdpartyresource.py
mward29: The string coming from Kubernetes can not be decoded (json.load) to a json object in its raw form. The json.dump encodes the string to a proper format so json.jump can decode it. I know it looks ackward but it works. It will not work without encoding it first.

lakshmi-kannan: I see. utf-8 goofup somewhere. It would good to note this as a comment for future contributors.

packs/kubernetes/sensors/thirdpartyresource.py
lakshmi-kannan: Let me try:

```
for line in lines:
   try:
      trigger_payload = self._get_trigger_payload_from_line(line)
   except MalformedObjectException:
      sys.exit(1)
   else:
      self._sensor_service.dispatch(trigger=trigger, payload=trigger_payload)
```

```
def _get_trigger_payload_from_line(self, line):
    line = self._fix_encoding(line)
    payload = self._build_trigger_payload_from_line(line)
    return payload
```

````
def _build_trigger_payload_from_line(self, line):
   # object building foo
```

Now you can unit test each of those functions.

lakshmi-kannan:
Getting better.

mward29:
lakshmi-kannan. I made your requested changes. At least I believe so. Let me know if my changes are what you had in mind.

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,103 @@
...
+            self._build_a_trigger(resource_type=resource_type, name=name, labels=labels_data,
+                                  namespace=namespace, object_kind=object_kind, uid=uid)
+        else:
+            self._log.debug("No Labels for the resource below. Tough to proceed without knowing how \
+                  to work with this object.")
+            self._log.debug(name, namespace, uid)
+            sys.exit(1)
+
+    def _build_a_trigger(self, resource_type, name, labels, namespace, object_kind, uid):
+        trigger = 'kubernetes.thirdpartyobject'
```
lakshmi-kannan: Remove this.

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,103 @@
...
+            'labels': labels,
+            'namespace': namespace,
+            'object_kind': object_kind,
+            'uid': uid
+        }
+
+        self._log.debug('Triggering Dispatch Now')
+
+        # Create dispatch trigger
+        self._sensor_service.dispatch(trigger=trigger, payload=payload)
```
lakshmi-kannan: This is not needed since you dispatch in line 39. It should be return payload.

packs/kubernetes/sensors/thirdpartyresource.py
```diff
@@ -0,0 +1,103 @@
...
+            uid = line['object']['metadata']['uid']
+        except:
+            self._log.debug("type, kind, name, namespace or uid do not exist in the object.\
+                              must exit")
+            sys.exit()
+
+        # Now lets see if labels exist, if so build a trigger else exit
+        if 'labels' in line['object']['metadata']:
+            labels_data = line['object']['metadata']['labels']
+            self._build_a_trigger(resource_type=resource_type, name=name, labels=labels_data,
```
lakshmi-kannan: return self._build_a_trigger(...)

PR #399, opened 2016-02-16
Typeform pack

emedvedev:
Transferring the Typeform pack from `st2incubator`. The pack is still a bit too specific and requires some fiddling with Typeform fields and a MySQL database to set the sensor up, but it’s still mature enough to list in `contrib`, I think. 

Removed the action chains though, those _are_ too specific (also depend on other packs). This is the reason I’m not removing the pack from incubator just yet: if we want to save the chains, we have to move them elsewhere first. `st2-community` maybe?

Will wait for someone’s LGTM then merge.


Kami:
I'm fine with that for now as long as all the tests / checks succeed.

PR #401, opened 2016-02-18
Use lint configs from lint-configs subtree repository 

Kami:
Part of https://github.com/StackStorm/st2/pull/2503#issuecomment-185323722 change.


Kami:
I also noticed there is a bug in the Makefile targets.

We don't correctly exit if a loop iteration returns non-zero (aka fails). The Makefile target would only fail if the last item in the loop iteration exited with non-zero.

This means there are some issues in the master branch which have managed to slip through because of this bug.

In any case, I'm working on a fix in this branch.

PR #402, opened 2016-02-18
Update affected action tests to use "get_action_instance" method to obtain Action class instance

Kami:
Part of this change - https://github.com/StackStorm/st2/pull/2511.


Kami:
Pushed some more goodies.

You can now force checks on a specific pack using ``FORCE_CHECK_PACK=<pack name> make all`` notation.

PR #405, opened 2016-02-19
VSphere pack

emedvedev:
Transferring from st2incubator


manasdk:
Will go wit https://github.com/StackStorm/st2contrib/pull/460 instead.

PR #413, opened 2016-02-23
Contribution guide

emedvedev:
A contribution guide and a PR template for packs.

cc @dzimine


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+## Pack (replace this with the pack's name)
```
Kami: Nice :+1:

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
+# Contributing to StackStorm pack directory
+
+You've decided to create or update a StackStorm integration pack. That's great!
```
Kami: As mentioned on Slack, I would prefer to move this to the documentation in the existing development docs section. Potentially this also requires some updates to the existing docs because we already have some docs on pack development. Then we can just link to it from this file.

This way most of the docs are located in a single place (less duplication) and harder for things to get out of sync.

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+
+* If you're contributing to a pack, please __update the version__ in `pack.yaml`! 
+  Without the version bump your PR will not be accepted. Pack versions follow 
+  [SemVer](http://semver.org/), and you should declare the new version 
+  accordingly.
+
+* If you're creating a brand-new pack, make sure all the __metadata__ is in place: 
+  the `pack.yaml` file, the necessary folder structure, ideally a 64x64 PNG icon.
+
+* It's important that all packs remain stable and reliable, so we recommend
```
Kami: Should also document `FORCE_CHECK_PACK=<pack name> make {all,lint}` command.

This command runs all the checks which run on Travis for a specific pack. Some of the checks require full st2 setup though since they connect to MongoDB.

emedvedev: It should probably go into https://docs.stackstorm.com/development/pack_testing.html if we're trying to move all "substance" to the docs. Wanna create a PR or should I?

CONTRIBUTING.md
emedvedev: The way I tried to keep it is having all "substance" (i.e. concrete examples, requirements, linting profiles, etc.) in the docs, and having a brief checklist in this document. We mostly need this file for quick reference, and the docs focus on "heavier" things like development, structure, meta, requirements, etc. This doc only contains a six-item checklist and a bunch of links to docs, and I tried to keep it as independent from the docs as possible: you'll always have to remember your meta, always have to update the version and format your code, no matter what the actual requirements for writing packs are.

So, I'd say we should merge this doc as it is (and keep it equally generic in the future), but also update st2docs to reflect some substantial things here: e.g. I don't think we have icon requirements in the docs, or a request for the version bump. Wdyt?

CONTRIBUTING.md
Kami: I want to put it here since it's st2contrib specific and pack tests itself are not st2contrib specific - you could also write tests for your pack or whatever where those targets don't exist.

emedvedev: Done! I wasn't sure whether to put it into formatting or testing, so I just created an additional section (a very small one though).

PR #422, opened 2016-03-02
Add BitBucket post_receive_webhook rule for autodeployment of commits

jjm:
### Status 

- pack extension, complete

### Description

- Add `rules/post_receive_webhook.yaml` which calls packs.deploy that was added in PR#????
- Add emacs editor saves to .gitignore
- Document new rule in README.md.

### DIFF
--- a/packs/bitbucket/README.md
+++ b/packs/bitbucket/README.md
@@ -193,7 +193,9 @@ This has a number of pre-dependancies:
 
 - Setting an Workflow / Hooks / Post-Receive WebHooks pointing at the URL
 
-```https://<my-server>/api/v1/webhooks/bitbucket_post_receive?st2-api-key=<ST2-API-Key```
+```
+https://<my-server>/api/v1/webhooks/bitbucket_post_receive?st2-api-key=<ST2-API-Key
+```
 
 *Important:* The BitBucket server (or cloud) needs to be able to reach
 your StackStorm server and considers the SSL cert as valid. The
--- a/packs/bitbucket/README.md
+++ b/packs/bitbucket/README.md
@@ -194,7 +194,7 @@ This has a number of pre-dependancies:
 - Setting an Workflow / Hooks / Post-Receive WebHooks pointing at the URL
 
 ```
-https://<my-server>/api/v1/webhooks/bitbucket_post_receive?st2-api-key=<ST2-API-Key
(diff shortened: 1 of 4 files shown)


Kami:
Thanks for the contribution.

packs/bitbucket/rules/post_receive_webhook.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: "post_receive_webhook"
+pack: "bitbucket"
+description: "Trigger ST2 pack deploy for commit on Git branch"
+enabled: true
```
Kami: It would be a good idea to disable this rule by default.

We disable rules which are part of common packs by default since they tend to be installation specific and user who wants to use it should explicitly enable it.

Kami: Also, this rule is a bit of a special case since it relies on functionality from two packs (there are many like that of course, but they tend to be installation specific).

We don't really have a good established pattern for it yet, but maybe it would be better if it's part of `packs` pack. In any case, I will also wait for others to chime in.

/cc @lakshmi-kannan @manasdk

jjm: I've disabled the rule as suggested. 

I think putting it in the packs pack does not make great sense to me as it would then be part of every install of StackStorm when it's only needed by those using BitBucket.

May be we need to pack-requirements.txt to document cross pack requirements?

emedvedev: I think having a dependency on default packs (chatops, st2, packs) is fine. +1 on merging.

packs/bitbucket/README.md
```diff
@@ -175,3 +175,29 @@ Usage:
 ```bash
 st2 run bitbucket.list_branches repo="<repo_name>"
 ```
+
+## Rules
+
+### Post-Receive WebHook
```
Kami: Can you please also document that this rule needs to be explicitly enabled (since it's disabled by default)?

Besides that, LGTM

jjm: A note to enable the rule (and the CLI command) has been added.

Kami:
Great, thanks.

I will go ahead and merge it into master.

PR #425, opened 2016-03-03
Fix bug in salt pack

tonybaloney:
### Status 

Replace this: describe the pack and PR status. Examples: 

- bugfix

### Description

Fixes a bug in the Salt pack, args are accepted in the action metadata but the request to API doesn't include the args in the payload. Updated the request logic to include args.

Also added 2 aliases for fun.


emedvedev:
Looks good, merging. Thanks! Also, aliases are always good. :stuck_out_tongue:

PR #431, opened 2016-03-03
[salt] Prevent premature args to list coersion

jcockhren:
### Status 

- bugfix

### Description

This PR fixes a issue introduced in #425 where `args` supplied to the action call where coerced to a `list` before checking if `args is None` (meaning it wasn't supplied with the action call). This causes:

1. failure in `local.py` because `list(args)` throws exception when `args` is `None`.
2. Supplying an empty `args` at during action call causes failure for any execution module that doesn't accept positional `args`

@tonybaloney @emedvedev


emedvedev:
Thanks for catching this! Will review shortly.

timoguin:
Looks good! :+1:

tonybaloney:
oops. sorry @jcockhren

PR #435, opened 2016-03-06
Fix an execution list bug in the `st2` pack

emedvedev:
The issue was reported via e-mail by Jonathon Adler.

The `st2` pack has a bug where trying to list pending executions would
result in an error.


Kami:
Kinda weird that `end_timestamp` is not there, it should always be there and just be `None` if the execution hasn't finished yet.

Did you confirm this change does indeed work?

emedvedev:
It does, I'm not sure why it wasn't there though.

Kami:
Weird, I'll check the code, since I would imagine something if `if exc.end_timestamp` would be more safe but it's possible that formatting code is doing something weird.

PR #438, opened 2016-03-08
Allow multiple exclamation marks while honking Tesla

emedvedev:
Since honking is a very emotional process, and people often have to
channel their intense feelings into it, it makes very little sense to
have a `?` quantifier in the honking alias: that would only allow for a
mere `honk-honk`, or a brief exclamative `honk-honk!`.

Changing the quantifier to `*` gives people the chance to express all
the feelings they want to express. Honk-honk!!!!!!!!!!!!!!!!!!!!!!!

### DIFF
--- a/packs/tesla/aliases/honk_horn.yaml
+++ b/packs/tesla/aliases/honk_horn.yaml
@@ -3,8 +3,11 @@ name: "tesla.honk_horn"
 action_ref: "tesla.honk_horn"
 description: "Honk your horn"
 formats:
-   - display: "honk-honk {{vin}}"
-     representation:
-       - "((beep|honk|meep)-?)+ {{vin}}[!.]*"
+    - display: "honk-honk"
+      representation:
+        - "((beep|honk|meep)-?)+[!.]*"
+    - display: "honk-honk {{vin}}"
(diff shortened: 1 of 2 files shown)


packs/tesla/aliases/honk_horn.yaml
```diff
@@ -5,6 +5,6 @@ description: "Honk your horn"
 formats:
    - display: "honk-honk {{vin}}"
      representation:
-       - "((beep|honk|meep)-?)+ {{vin}}[!.]?"
+       - "((beep|honk|meep)-?)+ {{vin}}[!.]*"
```
tonybaloney: vin is optional. this would mean you would have to put a space before the exclamation marks. 

`honk!!!!` wouldn't be accepted here

packs/tesla/pack.yaml
```diff
@@ -3,6 +3,6 @@ name : tesla
 description : car horn automation
 keywords:
   - tesla
-version : 0.1
+version : 0.1.1
```
tonybaloney: Since this is a major change to the functionality I would prefer 0.2.0 as the version number if we're following semver

PR #441, opened 2016-03-13
Update rules to use correct trigger name

rosskusler:
### Status 

Just a simple bugfix to make the trigger name referenced in the nagios rules match the trigger name that is actually registered.

### DIFF
--- a/packs/nagios/pack.yaml
+++ b/packs/nagios/pack.yaml
@@ -5,6 +5,6 @@ keywords:
   - nagios
   - monitoring
   - alerting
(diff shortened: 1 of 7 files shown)


emedvedev:
Awesome! Thanks for the catch.

Kami:
Good catch, indeed :)

Ideally, we would update the service handler so the full trigger name would be `nagios.service.state_change` to be consistent with other triggers (we tend to use `.` as a separator in the trigger name), but this approach also works for now.

PR #442, opened 2016-03-13
Add duo pack (2FA)

jjm:
### Status 

- new pack, Work in progress

### Description

A pack for to place 2FA authentication using the duo service (https://duo.com/) into workflows to add authentication / approval.


jjm:
This is mostly complete, just need to review the duo.auth action to ensure all the non-default options are covered.

Kami:
Great, thanks.

Will have a look shortly.

packs/duo/actions/auth.py
```diff
@@ -0,0 +1,60 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from st2actions.runners.pythonrunner import Action
+import duo_client
+
+
+class Auth(Action):
+    def run(self, username, factor, ipaddr, async, device,
```
tonybaloney: I wouldn't use a reserved word (`async`) as a parameter. It might come back and bite you in future

packs/duo/actions/auth.py
```diff
@@ -0,0 +1,60 @@
...
+                             username=username,
+                             type=push_type,
+                             device=device)
+        except:
+            raise ValueError("Duo Auth request failed!")
+        else:
+            if data['status'] == "allow":
+                return data
+            elif data['status'] == "deny":
+                print data['status_msg']
```
tonybaloney: Use the logging functions instead of printing to stdout

packs/duo/actions/auth.yaml
```diff
@@ -0,0 +1,55 @@
...
+parameters:
+  username:
+    type: "string"
+    description: "The Duo username to auth."
+    default: "{{action_context.api_user}}"
+
+  factor:
+    type: "string"
+    description: "Factor to use for authentication."
+    enum:
```
tonybaloney: I didn't even know you could do this!  Nice.

packs/duo/actions/auth.yaml
```diff
@@ -0,0 +1,55 @@
...
+  passcode:
+    type: "integer"
+    description: "Passcode entered by the user."
+
+  # FIX
+  push_info:
+    type: "string"
+    description: "*Fix* Extra key/value pairs that will be URL-encoded."
+
+runner_type: "python-script"
```
tonybaloney: Put this at the top, its really handy for the preview feature.

packs/duo/actions/ping.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+
+        try:
+            ikey = self.config['ikey']
+            skey = self.config['skey']
+            host = self.config['host']
+        except KeyError:
+            raise ValueError("Duo config not found in config.")
+
+        auth = duo_client.Auth(ikey=ikey,
```
tonybaloney: Seems you're using this method over and over in each action, you should prob create a `lib` submodule and place it in there, look at how the other packs do it. When you add more functions it'll save you refactoring later.

packs/duo/aliases/check.yaml
```diff
@@ -0,0 +1,18 @@
...
+name: "check"
+pack: "duo"
+description: "Test the intergration setup to the Duo platform"
+action_ref: "duo.check"
+formats:
+  - "duo check"
+ack:
+  enabled: true
+  append_url: false
+  format: "Chekcing setup for Duo Authication..."
```
tonybaloney: typo

packs/duo/aliases/auth.yaml
```diff
@@ -0,0 +1,19 @@
...
+name: "auth"
+pack: "duo"
+description: "Test your authication via the Duo platform"
+action_ref: "duo.auth"
+formats:
+  - "duo auth"
+ack:
+  enabled: true
+  append_url: false
+  format: "Spending a Duo push request..."
```
tonybaloney: Typo

packs/duo/aliases/auth.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: "auth"
+pack: "duo"
+description: "Test your authication via the Duo platform"
```
tonybaloney: typo

packs/duo/aliases/check.yaml
```diff
@@ -0,0 +1,18 @@
+---
+name: "check"
+pack: "duo"
+description: "Test the intergration setup to the Duo platform"
```
tonybaloney: typo

packs/duo/aliases/ping.yaml
```diff
@@ -0,0 +1,18 @@
...
+name: "ping"
+pack: "duo"
+description: "Check the Duo platform can be reached."
+action_ref: "duo.ping"
+formats:
+  - "duo ping"
+ack:
+  enabled: true
+  append_url: false
+  format: "Cheking connection to Duo..."
```
tonybaloney: typo

packs/duo/requirements.txt
```diff
@@ -0,0 +1 @@
+duo_client
```
tonybaloney: Prob better to specify the version you used. e.g. `duo_client==0.1.2`

packs/duo/actions/auth.py
jjm: However I'm attempting to get some human status text (status_msg) from Duo out of the runner that can be shown to the user in the alias without extras, 

From [st2common/log.py](https://github.com/StackStorm/st2/blob/master/st2common/st2common/log.py) all logging methods are decorated and sent to stderror.

When an action succeeds we can `return` a dict and access the data in the result in the alias or publish in a workflow. However if we need an action to fail we don't seem to return data. So this is the first comprise that I've come up with.

May be the action should should just return the data from Duo and leave the processing of the status to the workflow it would be contained in. And it only fail if there was some underling error. However the succeed for being authenticated and failing for being denied would a visual representation in flow.

packs/duo/actions/auth.yaml
jjm: It's useful, until you put the action into a Mistral workflow flow where it tracebacks as `{{action_context.api_user}}` is not defined.

jjm:
The auth side of things is basically covered, the approach for handling output to the user when the action user is denied access by Duo needs to be consider further.

Options:
- have action succeed on allow & deny and expect the workflow to parse the response to know if the auth was successful.
- Only succeed if auth is allowed and return data and print message + raise an exception when auth is denied.

Personally I think the failure of the action on `deny` would be easier to understand instead having to put the following into the Mistral workflow: 

```yaml
on-successs:
   - run_action: <% $.status = "allow" %>
   - no_auth: <% $.status = "deny" %>
```

[The above is hidden in a comment on an out dated diff]

emedvedev:
Awesome, I'll try with my Yubikey some time later.

jjm:
@emedvedev Sorry, all my testing has been with `factor` set to `auto` so far, which translates to `push` or `phone`. Totally lacking the code in `auth_auth.py` to handle the other factors or pass in the passcode so far.

Before it gets to the merging stage I'll need to make that function, but there would need to be a action to pick up the passcode from another location and pass it to the `duo.auth_auth` action.

jjm:
@emedvedev I've made it handle the different factors correctly now. it's worked for me with a passcode from the app so a Yubikey shoujld be fine now.

jjm:
Latest commit has addressed the lack of lib use by @tonybaloney, so if this passes CI checks I think we are good to merge.

packs/duo/actions/auth_auth.yaml
```diff
@@ -0,0 +1,46 @@
...
+    type: "string"
+    description: "This string is displayed in the Duo Mobile app before the word “request”. "
+    default: "ChatOps"
+
+  passcode:
+    type: "string"
+    description: "Passcode entered by the user."
+
+  pushinfo:
+    type: "string"
```
Kami: It seems like the type of this argument should be `object`.

This way you don't need need to do any funky splitting in the code.

packs/duo/actions/auth_auth.py
```diff
@@ -0,0 +1,82 @@
...
+                auth_kargs['pushinfo'] = encoded
+        elif factor == "passcode":
+            auth_kargs['passcode'] = passcode
+        elif factor == "phone":
+            auth_kargs['device'] = device
+        elif factor == "sms":
+            # As 'sms' just denies and then we do not support it
+            # requires re-authentication.
+
+            print "Denied, we do not support SMS!"
```
Kami: Please use `self.logger.info` here.

packs/duo/actions/auth_auth.py
```diff
@@ -0,0 +1,82 @@
...
+            raise ValueError("Denied, we do not support SMS!")
+        else:
+            raise ValueError("Invalid factor!")
+
+        try:
+            data = self.duo_auth.auth(factor=factor,
+                                      username=username,
+                                      **auth_kargs)
+        except RuntimeError, e:
+            print "Error: %s" % e
```
Kami: `self.logger.error` please

packs/duo/actions/auth_auth.py
```diff
@@ -0,0 +1,82 @@
...
+            raise RuntimeError("Error: %s" % e)
+        else:
+            if data['result'] == "allow":
+                return data
```
Kami: Maybe some logging - e.g. Authentication succeed or similar.

packs/duo/actions/auth_ping.py
```diff
@@ -0,0 +1,39 @@
...
+        Returns: An dict with info returned by Duo.
+
+        Raises:
+          RuntimeError: On ping failure.
+        """
+
+        try:
+            data = self.duo_auth.ping()
+        except RuntimeError, e:
+            print "Ping failed! %s" % e
```
Kami: Please use logger :)

Kami:
Thanks!

I added a couple of more comments.

jjm:
Thanks to everyone for the comments and feedback, the pack covers the 3 useful actions for use of Duo within ST2. So I think we are good to merge, at a later date I expect I'll be able to start covering the admin API too.

Kami:
Merged, thanks!

PR #443, opened 2016-03-15
Fix pip typo in `duo` requirements

emedvedev:
Totally overlooked that. :) The pack won’t install without the fix because the syntax is incorrect.


Kami:
:+1:

PR #444, opened 2016-03-15
Hubot-stackstorm native support for Duo

emedvedev:
Enhancing the Duo pack with support for the corresponding `hubot-stackstorm` change, making two-factor auth an `extra` property in an alias:

```
# cat aliases/repeat.yaml 
---
name: "echo"
description: "Repeat a sentence"
action_ref: "chatops-training.repeat"
formats:
[…]
ack:
  append_url: false
  format: "Ack! I'll repeat your sentence now."
result:
  format: "here goes: {~} {{ execution.result.stdout | truncate(40) }}"
extra:
  security:
    twofactor: true
```

The extra alias sends a Duo request and then either makes a `2fa` announcement which `hubot-stackstorm` will catch or posts a failure report.


emedvedev:
screenshot 2016-03-16 00 37 32

emedvedev:
PR on the `hubot-stackstorm` side: https://github.com/StackStorm/hubot-stackstorm/pull/102

PR #447, opened 2016-03-17
Update documentation on how to use instance IAM roles and set Defaults

ravenbyron:
### Status 

- Documentation update and set defaults

### Description

This updates the documentation to allow the AWS pack to use instance IAM role and also sets the defaults in the config.yaml to use US-EAST-1 and Instance IAM role for connections, this will allow it to work out of the box for anyone in running in US-EAST-1 with a meaningful role assigned. 

- [ *] Version bump (required for changed packs)

### DIFF
--- a/packs/aws/config.yaml
+++ b/packs/aws/config.yaml
@@ -1,8 +1,8 @@
 ---
 setup:
-  region: ""
-  aws_access_key_id: ""
-  aws_secret_access_key: ""
+  region: "us-east-1"
+  aws_access_key_id: null
+  aws_secret_access_key: null
 interval: 20
 st2_user_data: "/opt/stackstorm/packs/aws/actions/scripts/bootstrap_user.sh"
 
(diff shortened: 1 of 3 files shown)


timoguin:
I like it.

Kami:
Thanks.

The PR looks good to me.

/cc @DoriftoShoes

packs/aws/config.yaml
```diff
@@ -1,8 +1,8 @@
 ---
 setup:
-  region: ""
-  aws_access_key_id: ""
-  aws_secret_access_key: ""
+  region: "us-east-1"
```
manasdk: No sure if you intended to check-in `us-east-1`. Typically the config should be empty unless the default value is unambiguous :)

manasdk:
+1

packs/aws/config.yaml
ravenbyron: Not sure what the standard is here but was trying to go for something that worked for many people out of the box...   I can change if needed

manasdk: In my view safer to not have a value specified and have anyone using the pack make an explicit choice.

ravenbyron: Changed Back

PR #448, opened 2016-03-17
Add Microsoft SQL Server Pack (MSSQL)

codyaray:
### Status 

- new pack, complete

### Description

Pack which allows integration with Microsoft SQL Server.

```
Action                      | Description                                     | Example or Use Case
----------------------------| ----------------------------------------------------------------------------------
**execute.scalar**          | Returns first column of first row from result.  | `SELECT COUNT(*) FROM employees`
**execute.row**             | Returns first row of data from result.          | `SELECT * FROM employees WHERE id=13`
**execute.insert**          | Returns newly inserted row identity.            | `INSERT INTO employees VALUES ('Cody', 'Ray')`
**execute.non_query**       | Returns number of affected rows.                | `INSERT`, `UPDATE`, `DELETE`, or DDL commands
**execute.query**           | Returns list of CSV files containing result.    | `SELECT` or `EXEC` stored procedures
**execute.query_and_email** | Emails all query results as attached files      | `SELECT` or `EXEC` stored procedures
```

- [X] Tests (not required but really recommended)


manasdk:
Nice!

codyaray:
Not sure how ti fix these pylint errors.

```
************* Module lib.mssql_action
E:  1, 0: No name 'connect' in module '_mssql' (no-name-in-module)
************* Module lib.results_processor
E:  1, 0: No name 'ROW_FORMAT_DICT' in module '_mssql' (no-name-in-module)
************* Module mssql_runner
E:  1, 0: No name 'MSSQLException' in module '_mssql' (no-name-in-module)
make: *** [.pylint] Error 1
```

Locally I tried fixing this by replacing the `from _mssql import X` with `import _mssql` and using the namespaced call, but I get a related error then:

```
Running pylint on pack: mssql
************* Module mssql_runner
E: 35,15: Module '_mssql' has no 'MSSQLException' member (no-member)
************* Module lib.results_processor
E: 76,49: Module '_mssql' has no 'ROW_FORMAT_DICT' member (no-member)
************* Module lib.mssql_action
E: 17,15: Module '_mssql' has no 'connect' member (no-member)
make: *** [.pylint] Error 1
```

These things are in the module. Is it looking for an `__all__` or something? This is third-party code, so not sure what else to do.

Kami:
Awesome work.

So it looks like the apt-package .travis.yaml changes I pushed to master helped with lint issues?

packs/mssql/actions/execute.query.yaml
```diff
@@ -0,0 +1,33 @@
...
+    default: execute_query
+    immutable: true
+    type: string
+  query_string:
+    description: Query to send to MS SQL Server.
+    required: true
+    type: string
+  params:
+    description: Parameters for interpolation in the query_string.
+    type: string
```
Kami: It would be better for `params` to be an object instead of a string (ideally you should always try to utilize a native type).

This way you don't need to run potentially unsafe `ast.literal_eval` in the code since `params` will already be a dictionary.

packs/mssql/actions/execute.query_and_email.yaml
```diff
@@ -0,0 +1,44 @@
+---
+name: "execute.query_and_email"
+description: "Executes an MS SQL Query and emails results as attachments"
+enabled: true
+
+runner_type: "action-chain"
+entry_point: "workflows/execute.query_and_email.yaml"
+
+parameters:
+  query_string:
```
Kami: It looks like this action (unlike others) doesn't support parameter expansion and you need to provide all the parameters in-line in the `query_string` parameter.

Is this correct?

Kami: Please ignore the comment above.

I some how missed `params` parameter in the metadata file which is defined just below (lol).

packs/mssql/actions/lib/mssql_action.py
```diff
@@ -0,0 +1,32 @@
...
+        db_config = self.config.get(database, {})
+        params = {
+            'database': db_config.get('database') or database,
+            'server': server or db_config.get('server'),
+            'user': user or db_config.get('user'),
+            'password': password or db_config.get('password')
+        }
+        unspecified = [param for param, value in params.iteritems() if value is None]
+        if unspecified:
+            raise Exception("Must specify or configure in config.yaml: %s" % ', '.join(unspecified))
```
Kami: It seems like a better message might be something along the lines of `Missing value for the required params: param1,param2` since if I understand it correctly, `params` are provided as action parameter and not defined in the config.

packs/mssql/actions/lib/mssql_action.py
```diff
@@ -0,0 +1,32 @@
...
+from st2actions.runners.pythonrunner import Action
+
+__all__ = [
+    'MSSQLAction'
+]
+
+
+class MSSQLAction(Action):
+
+    def __init__(self, config):
```
Kami: You don't actually need __init__ method here, __init__ method of the parent class already does that so you can simply get rid of it here.

packs/mssql/actions/lib/results_processor.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, config, logger):
+        self.config = config
+        self.logger = logger
+
+    def execute_scalar(self, response, cursor):
+        """
+        Returns the scalar response.
+        """
+        if not response:
+            sys.exit(self.NO_DATA)
```
Kami: You could also throw an exception here - this will also cause action to fail, but it will make it more obvious to the user what is going on because exception can contain an error message.

packs/mssql/actions/lib/results_processor.py
```diff
@@ -0,0 +1,104 @@
...
+        Returns identity value of the inserted row.
+        """
+        return cursor.identity
+
+    def execute_non_query(self, response, cursor):
+        """
+        Returns the number of rows affected by the non-query.
+        """
+        if cursor.rows_affected == 0:
+            sys.exit(self.NO_DATA)
```
Kami: Same here.

packs/mssql/actions/lib/results_processor.py
```diff
@@ -0,0 +1,104 @@
...
+                    os.unlink(csv_file.name)
+                    # we're done, no more result sets
+                    break
+                output_files.append(csv_file.name)
+                writer = csv.DictWriter(csv_file, fieldnames=first_row.keys())
+                writer.writeheader()
+                writer.writerow(first_row)
+                for row in cursor:
+                    writer.writerow(self._filter_numbered_columns(row))
+        if not output_files:
```
Kami: Should no results always be a failure? 

I think that is somewhat use-case specific and in some cases no results / rows simply indicates no data which is not necessary a failure.

Kami:
I've added some comments.

Overall it looks good and besides the `params` parameter type, it's mostly minor things :)

packs/mssql/actions/execute.query.yaml
codyaray: I think that I actually started with an object. We lose some flexibility by using an object though, unless I just preprocess the object back into a format like this.

The README might make this make more sense.

```python
    query_string = "SELECT * FROM empl WHERE id=%d"
    params = 13
    
    query_string = "SELECT * FROM empl WHERE name=%s"
    params = "'John Doe'"
    
    query_string = "SELECT * FROM empl WHERE name LIKE %s"
    params = "'J%'"
    
    query_string = "SELECT * FROM empl WHERE name=%(name)s AND city=%(city)s"
    params = "{ 'name': 'John Doe', 'city': 'Nowhere' }"
    
    query_string = "SELECT * FROM cust WHERE salesrep=%s AND id IN (%s)"
    params = "('John Doe', (1, 2, 3)))"
```

For what its worth, `ast.literal_eval` [is always safe](https://docs.python.org/2/library/ast.html#ast.literal_eval), unlike `eval`.

packs/mssql/actions/lib/results_processor.py
codyaray: Yeah, I was hoping someone would bring this usage point up. The current is modeled after `grep`, which returns a failure exit code when it doesn't find a match. My original thinking was that it would make decisions in workflows easier.

For example, in the `execute.query_and_email` action, we specifically don't want to email if there's `NO_DATA` returned. Of course, we don't want this to show up as a failure though, so we end up adding an extra task in the workflow anyway. 

```yaml
  -
    name: 'query_mssql'
    ref: 'mssql.execute.query'
    parameters:
      query_string: '{{query_string}}'
      params: '{{params}}'
      database: '{{database}}'
    # success indicates returned data. don't email if NO DATA
    on-success: 'email_results'
    # failure may indicate no data or an actual error
    on-failure: 'success_on_no_data'
 ```

I guess we either go from on-success into a "router" action which decides the next action based on whether it `HAS_DATA` or `NO_DATA`, or we go from on-failure into a similar action which decides based on `NO_DATA` or `HAS_ERROR`. In other words, we have three possible states but only two outputs, so either way isn't super clean.

Do you have any better thoughts, suggestions, or patterns here? For example, should failure always indicate an exception, or can it be used for conveying another meaning (e.g., `NO_DATA`)?

Comment (Kami) [inline packs/mssql/actions/lib/results_processor.py]: Preferred option would be to only throw and fail on actual failure and determine if the action returned any data by inspecting the result object inside the workflow (e.g. if len result == 0 or similar).

If it makes it easier, you could also make the action return an object which includes additional key such as `no_data` or similar.

File: packs/mssql/actions/execute.query.yaml (hunk 1 above)
Comment (Kami) [inline packs/mssql/actions/execute.query.yaml]: I see, thanks for the clarification.

In this case I'm fine with making an exception and using a string for the parameter type here.

**Comment by Kami** [comment]
I will go ahead and merge this into master, but I would appreciate if you could follow and and either change the sys.exit thing to an exception or return an empty list or similar when there are no results.

Thanks again!

File: packs/mssql/actions/lib/results_processor.py (hunk 7 above)
Comment (codyaray) [inline packs/mssql/actions/lib/results_processor.py]: I can return successful regardless of whether data is returned. Looks like we'll need to add an extra task to the workflow that'll check to send an email only if data is returned and show the workflow as successful even if no data is returned.

```yaml
  -
    name: 'check_for_results'
    ref: 'core.local'
    parameters:
      # mssql.execute.query returns 1 on ERROR and 2 if NO DATA
      cmd: '[[ "{{query_mssql.result.output_files | length}}" -gt 0 ]]'
    on-success: 'email_results'
    on-failure: 'success_on_no_data'
```

Do you see a shorter or better way?

codyaray: Okay, figured it out. We can use sendmail's `send_empty_body` to conditionally send.

New PR for MSSQL updates: #455 
PR for improved sendmail output when not sending anything: [#2597](https://github.com/StackStorm/st2/pull/2597)

PR #450, opened 2016-03-18
use *array* for cloudformation parameters and capabilities.

pidah:
### Status  
- bugfix

### Description
Cloudformation create_stack function expects a list as shown here:
https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/cloudformation/connection.py#L157
https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/cloudformation/connection.py#L341

### DIFF
--- a/packs/aws/actions/cf_create_stack.yaml
+++ b/packs/aws/actions/cf_create_stack.yaml
@@ -8,7 +8,7 @@ parameters:
     immutable: true
     type: string
   capabilities:
-    type: string
+    type: array
   cls:
     default: CloudFormationConnection
     immutable: true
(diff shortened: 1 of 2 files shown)


Kami:
Good catch, thanks.

I'm actually not sure how this will affect our meta generator code, since IIRC, most of those metadata files are automatically generated.

/cc @DoriftoShoes

pidah:
@kaml thanks.  I didn't know they were auto-generated. Anyway, this was the only way I could get it working.

manasdk:
@pidah Thanks for the change!

@Kami This is fine. We have to make these exceptions. I plan to rewrite the auto-gen scripts to look up types from the python-docs and until then this is the only way to fix things where type cannot be inferred from default values.

PR #453, opened 2016-03-23
[salt] Hotfix/missing target args

jcockhren:
### Status 

- bugfix

### Description

This bugfix ensures the developer can customize which minons are targeted and how when using [Execution Module Actions](https://github.com/StackStorm/st2contrib/tree/master/packs/salt#execution-module-actions)

### DIFF
--- a/packs/salt/actions/local_archive.gunzip.yaml
+++ b/packs/salt/actions/local_archive.gunzip.yaml
@@ -11,7 +11,7 @@ parameters:
   target:
      required: false
      type: string
-     default: *
+     default: '*'
   expr_form:
     required: false
     type: string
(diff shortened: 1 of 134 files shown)


Kami:
Thanks for the contribution.

Just curious - was this argument added to ansible in one of the recent versions?

Kami:
It looks like salt calls parameter `tgt` for some reason.

Was there some reason why you used `target` instead of `tgt`? Using `tgt` might make more sense if it keeps it consistent with the Salt parameter names.

jcockhren:
@Kami yeah, that was on purpose. I wanted to use `target` since it's clearer.

Kami:
Alright, as long as this is obvious to the user I'm fine with that.

PR #456, opened 2016-03-24
Fix MSSQL README actions table formatting

codyaray:
Realized that Github doesn't like my markdown for the actions table.
https://github.com/StackStorm/st2contrib/tree/master/packs/mssql

Fix it so it works locally and on Github.


emedvedev:
Thanks!

PR #462, opened 2016-04-01
Vsphere barebones atomic actions

paul-mulvihill:
### Status 

- pack extension

### Description

Creation of atomic acitons allowing or basic VM creation from scratch (non templated). Actions/flows are prefixed with vm_hw to signify they are targetting the VM hardware not OS interaction via vmtools.
These actions are intented to create a VM, power it on and then allow a build server to take over OS deployment.


andrew-regan:
Pretty sure pyVmomi v6 changed the way SSL certificate verification is handled. If you want to use a self-signed certificate with v6 you need to pass a SSL context into `connect.SmartConnect()` in `st2contrib/packs/vsphere/actions/vmwarelib/actions.py`.

For an example take a look at this code from https://github.com/dnaeon/py-vconnector/blob/master/src/vconnector/core.py:
```
        if not ssl_context:
            ctx = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
            ctx.verify_mode = ssl.CERT_NONE
            self.ssl_context = ctx
        else:
            self.ssl_context = ssl_context
```
...
```
            self._si = pyVim.connect.SmartConnect(
                host=self.host,
                user=self.user,
                pwd=self.pwd,
                sslContext=self.ssl_context,
            )
```

paul-mulvihill:
Hi Andrew.

I had seen mention of that in the past. Although if i understand things right I needs a minimum of python 2.7.9, because of the ssl module. Which isn't available via apt-get on ubuntu (2.7.6 is deamed latest, at least by my ubuntu server).
I want to eventually have our config management system build out st2 instances and therefore don't really want to deviate from what is available by the standard avenues for package installation.
I may be missing some info but for later python releases don't you have to deploy from source?

Paul

andrew-regan:
I'm not sure on the 2.7.9 requirement, but it could be working for us because we're using StackStorm w/ Python 2.7.11 built from source on CentOS.

packs/vsphere/actions/vm_hw_barebones_create.py
```diff
@@ -0,0 +1,87 @@
+from pyVmomi import vim
```
manasdk: Please add a license header

packs/vsphere/actions/vm_hw_barebones_create.yaml
```diff
@@ -0,0 +1,52 @@
...
+      description: "Number of CPUs"
+      required: false
+      default: 1
+    ram_size:
+      type: "integer"
+      description: "Ammount of RAM in GB"
+      required: false
+      default: 2
+    guestos:
+      type: "string"
```
manasdk: This is likely an ENUM - right?

packs/vsphere/actions/vm_hw_cpu_mem_flex.yaml
```diff
@@ -0,0 +1,23 @@
+---
+  name: "vm_hw_cpu_mem_flex"
```
manasdk: Unclear on the `flex` suffix.

packs/vsphere/actions/vm_hw_barebones_create.py
```diff
@@ -0,0 +1,87 @@
...
+                                   memoryHotAddEnabled=True,
+                                   annotation=description)
+
+        # if Datastore cluster is provided it will find the
+        # recommended Datastore to store VM files
+        if datastore_cluster:
+            podsel = vim.storageDrs.PodSelectionSpec()
+            podsel.storagePod = data_store_cluster
+
+            storageSpec = vim.storageDrs.StoragePlacementSpec(
```
manasdk: suggest `storage_spec`

packs/vsphere/actions/vm_hw_barebones_create.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            # No Datastore Cluster has been offered so using the D
+            if datastore:
+                #data_store = inventory.get_datastore(self.si_content,
+                                                     #name=datastore)
+                datastore_path = '[' + datastore + '] ' + vm_name
+            else:
+                raise Exception('Error No Storage Data Provided')
+
+        #Now Datastore is known the remaining
```
manasdk: linter will not like this. space between # and comment is needed.

packs/vsphere/actions/vm_hw_barebones_create.py
```diff
@@ -0,0 +1,87 @@
...
+                                   vmPathName=datastore_path)
+        config.files = vmx_file
+
+        #Create task to Build actual Machine
+        task = target_folder.CreateVM_Task(config=config, pool=resource_pool)
+        self._wait_for_task(task)
+        if task.info.state != vim.TaskInfo.State.success:
+            raise Exception(task.info.error.msg)
+
+        return {'task_id': task._moId, 'vm_id': task.info.result._moId}
```
manasdk: Since you anyway wait for task returning the taskid likely will not provide much value.

packs/vsphere/actions/vm_hw_basic_build.yaml
```diff
@@ -0,0 +1,60 @@
...
+  vm_name: 
+    required: true
+    type: string 
+  vm_hdd_dscluster: 
+    type: string
+    description: Which Datastore Cluster should be used for this HDD. Blank = same as Base VM    
+  vm_resourcepool: 
+    type: string
+  vm_version: 
+    default: vmx-09
```
manasdk: you use `vmx-10` as a default in the WF. Likely want to use the same default.

packs/vsphere/actions/vm_hw_cpu_mem_flex.yaml
```diff
@@ -0,0 +1,23 @@
...
+  runner_type: "run-python"
+  description: "Adjust CPU and RAM allocation for a Virtual Machine"
+  enabled: true
+  entry_point: "vm_hw_cpu_mem_flex.py"
+  parameters:
+    vm_id:
+      type: "string"
+      description: "VM to Flex."
+      required: false
+    vm_name:
```
manasdk: I would suggest only using vm_id. If I remember right vm names are unique within some containers.

packs/vsphere/actions/vm_hw_details_get.py
```diff
@@ -0,0 +1,12 @@
+from vmwarelib import inventory
+from vmwarelib.actions import BaseAction
+
+
+class GetVMDetails(BaseAction):
+    def run(self, vms):
+        results = []
+        for vm in vms:
+            vm = inventory.get_virtualmachine(self.si_content, name=vm)
```
manasdk: This is ok for now but using a propertyspec to retrieve all at the same time will be better. Please leave a TODO for future updates.

packs/vsphere/actions/vm_hw_details_get.yaml
```diff
@@ -0,0 +1,10 @@
+---
+  name: vm_detail_get
+  runner_type: run-python
+  description: Retrieve details of a VM object
+  entry_point: vm_hw_details_get.py
+  parameters:
+    vms:
```
manasdk: feel these should be vm_ids else it will not be unique.

packs/vsphere/actions/vm_hw_nic_edit.py
```diff
@@ -0,0 +1,54 @@
...
+            raise Exception('Unable to find Network Adapter provided')
+
+        # Create object for new Specification
+        new_spec = vim.vm.device.VirtualDeviceSpec()
+        new_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
+        new_spec.device = nic
+
+        #If network name provided assign new network
+        # Room to expand the following to set additional flags/values
+        if network_name:
```
manasdk: Do you check this here instead of earlier because you plan to expand this action later and perform different edits?

packs/vsphere/actions/vm_hw_nic_edit.yaml
```diff
@@ -0,0 +1,24 @@
+---
+  name: "vm_hw_nic_edit"
```
manasdk: Will it ever do more than switching the network on a nic?

packs/vsphere/actions/vm_hw_power.py
```diff
@@ -0,0 +1,25 @@
+import eventlet
+from pyVmomi import vim
+
+from vmwarelib import inventory
+from vmwarelib import checkinputs
+from vmwarelib.actions import BaseAction
+
+
+class VMPowerOn(BaseAction):
```
manasdk: Name is misleading. Should be `VMApplyPowerState`

packs/vsphere/actions/vm_hw_power.py
```diff
@@ -0,0 +1,25 @@
...
+        # check I have information to find a VM
+        checkinputs.vm_identifier(vm_id, vm_name)
+        # convert ids to stubs
+        vm = inventory.get_virtualmachine(self.si_content,
+                                          moid=vm_id, name=vm_name)
+        if not vm:
+            raise Exception('Error: Unable to find VM')
+        if power_onoff == "poweroff":
+            task = vm.PowerOffVM_Task()
+        else:
```
manasdk: elif power_onoff = "poweron" to make it definite and not allow arbitrary strings

packs/vsphere/actions/vm_hw_power.yaml
```diff
@@ -0,0 +1,23 @@
+---
+  name: "vm_hw_power"
```
manasdk: In this case I prefer 2 separate action. vm_power_on and vm_power_off. Its more obvious from a usage standpoint.

packs/vsphere/actions/vmwarelib/checkinputs.py
```diff
@@ -0,0 +1,10 @@
+def vm_identifier(moid=None, name=None):
+    if ((name is None or name == "")
```
manasdk: Perhaps something like

```
if (name and isinstance(name, six.string_type)) or (moid and isinstance(moid, six.string_type)):
    return
raise Exception(...) 
```

This guarantees string types as well

packs/vsphere/actions/vmwarelib/inventory.py
```diff
@@ -1,23 +1,9 @@
-# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
manasdk: Any reason you dropped the license?

packs/vsphere/actions/vmwarelib/inventory.py
```diff
@@ -25,30 +11,73 @@ def get_managed_entity(content, vimtype, moid=None, name=None):
         elif name and entity.name == name:
             return entity
 
+    #if this area is reached no object has been found
+    #if no arguments were passed just return with empty object
+    if ((name is None or name == "") and (moid is None or moid == "")):
```
manasdk: `if not name and not moid:` is more "pythonic". Checking for falsy does have its pitfalls.

packs/vsphere/actions/vmwarelib/inventory.py
```diff
@@ -25,30 +11,73 @@ def get_managed_entity(content, vimtype, moid=None, name=None):
...
+        return
+    #if a name was passed error
+    elif (name is not None and name != ""):
```
manasdk: `if name:`

packs/vsphere/actions/vmwarelib/inventory.py
```diff
@@ -25,30 +11,73 @@ def get_managed_entity(content, vimtype, moid=None, name=None):
...
+        raise Exception("Inventory Error: Unable to Find Object (%s): %s"
+                        % (vimtype, name))
+    #if a moid was passed error
+    elif (moid is not None and moid != ""):
```
manasdk: `elif moid`

packs/vsphere/actions/vmwarelib/inventory.py
```diff
@@ -25,30 +11,73 @@ def get_managed_entity(content, vimtype, moid=None, name=None):
...
+        raise Exception("Inventory Error: Unable to Find Object (%s): %s"
+                        % (vimtype, moid))
+    #catch all error
+    else:
+        raise Exception("Inventory Error: No Name or moid provided (%s)"
```
manasdk: This should be a pre-condition and not at the end.

packs/vsphere/actions/workflows/vm_hw_basic_build.yaml
```diff
@@ -0,0 +1,69 @@
+---
```
manasdk: I like this approach of VM creation i.e. using a workflow. It might end up being less efficient as every task is an API call against vsphere however there is immense clarity and pluggability which IMO trumps the cost of API calls.

packs/vsphere/actions/vm_hw_barebones_create.yaml
paul-mulvihill: Yes cpu/ram/guestos could be enums, but that does also mean i'm limiting a input.
cpu/ram can be as high as vsphere/hardware can offer which may vary setup to setup.
guestos will change with each release of vsphere.

packs/vsphere/actions/vm_hw_cpu_mem_flex.yaml
paul-mulvihill: flex is just the same as edit (increase or decrease). I can rename as edit if that's better.

packs/vsphere/actions/vm_hw_nic_edit.yaml
paul-mulvihill: yes.. the aim of the action was to be a starting point. eventually it will be able to do more. for now i just need a network change action. No reason it can't be expanded to allow the update of other configuration information about the nic.

packs/vsphere/actions/vmwarelib/inventory.py
paul-mulvihill: the version I originally worked from must have not had it.. but i'll get it added back in.

packs/vsphere/actions/vm_hw_barebones_create.py
paul-mulvihill: I've been using flake8 and it's cleared all the python files. do i need to use something else as well?

packs/vsphere/actions/vm_hw_barebones_create.py
paul-mulvihill: will get removed.

packs/vsphere/actions/vm_hw_basic_build.yaml
paul-mulvihill: I left V10 in the action because it was the latest. in the basic build i put it as 9 as i'm using the desktop client for vsphere and it doesn't allow edit's on v10 vm's through the desktop app.

packs/vsphere/actions/vm_hw_nic_edit.py
paul-mulvihill: the if network_name line is exactly for that purpose.. none of my actions are intended as absolute.. i hope they will be expanded in the furture. The if network_name was intended to only update what is needed in the spec if someone passes the value.

packs/vsphere/actions/vm_hw_power.py
paul-mulvihill: sorry missed that.. i'll get it updated.

packs/vsphere/actions/vm_hw_cpu_mem_flex.yaml
paul-mulvihill: I'm hesitant to restrict this just to the vm id.. mainly as that means what ever system/person is calling this must know that value otherwise i can't be used. As an interim i've added in extra logic so if multiple items are found matching name or ID it will raise an exception.

packs/vsphere/actions/vm_hw_barebones_create.py
paul-mulvihill: done

Kami:
@paul-mulvihill #467 should fix the "weird" test issues you were seeing.

I will merge it into master once tests pass and then you can also merge master into your branch to get that change.

packs/vsphere/actions/vm_hw_power.py
```diff
@@ -0,0 +1,42 @@
...
+class VMApplyPowerState(BaseAction):
+
+    def run(self, vm_id, vm_name, power_onoff):
+        # check I have information to find a VM
+        checkinputs.one_of_two_strings(vm_id, vm_name, "ID or Name")
+        # convert ids to stubs
+        vm = inventory.get_virtualmachine(self.si_content,
+                                          moid=vm_id, name=vm_name)
+        if not vm:
+            raise Exception('Error: Unable to find VM')
```
jjm: Is `Exception` the best to raise, would `ValueError` be a better exception?

jjm:
And passing checks now!

Would it be an idea to include `CONTRIBUTORS.md` & `CHANGELOG.md` files to this pack too?

Kami:
@jjm Yeah, feel free to add `CONTRIBUTORS.md` file.

You can use a somewhat common format of `Firstname Lastname <email>` per line.

packs/vsphere/actions/vm_hw_power.py
Kami: Yeah, I think it's fine since `ValueError` refers to function / method arguments (e.g. argument is of a right type but contains an invalid / unsupported value) and not the result.

packs/vsphere/actions/vm_hw_barebones_create.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python
```
manasdk: Don't need that here since its a module and not a script to be executed in a shell.

packs/vsphere/actions/vm_hw_cpu_mem_flex.yaml
manasdk: Ok. Then lets go with edit.

manasdk:
Yes `contributors.md` makes sense.

I am +1 to this change.

paul-mulvihill:
Think i've handled any technical comments. Anything else you think needs resolving before we consider merging this?

PR #464, opened 2016-04-03
Tesla pack: add a Tesla 3 warning

emedvedev:
@tonybaloney: we absolutely need to test the Tesla pack for compatibility with Tesla 3 as soon as you can get your hands on one.

This is of utmost importance: we don’t want a flood of angry Tesla 3 customers on the release date saying they can’t honk their horn through Slack. If possible, please request a clarification from Elon on whether the current API will stay the same for Tesla 3.


manasdk:
Unless we get to be beta testers, mind you this is absolutely a valid reason, we might have to wait until end of 2017 :P

tonybaloney:
I just got off the phone with Elon, he said they outsourced the API development to IBM and they're going to have some JSON-over-XML transport layer https://www.ibm.com/support/knowledgecenter/SS9H2Y_7.1.0/com.ibm.dp.doc/json_jsonx.html unfortunately, this is developed in Java, so we then need to have a Struts mediator to represent the data as CSV and then call that controller from Python. 
Sadly, I won't be around to see this through, so I'll leave it in the capable hands of @emedvedev

PR #466, opened 2016-04-05
Add Solarwinds Orion pack.

jjm:
### Status 

Almost complete, just review comments and testing left.

- new pack, work in progress

### Description

Add pack for integrating with the Solarwinds Orion monitoring platform.

- [x] Tests (not required but really recommended)
- [X] Add more actions & alaises.
- [x] Refactor actions to use lib.
- [x] Pack installs from GitHub.


Kami:
Great, thanks for the PR.

I will review it shortly.

packs/orion/actions/lib/actions.py
```diff
@@ -0,0 +1,108 @@
...
+
+from st2actions.runners.pythonrunner import Action
+from orionsdk import SwisClient
+
+
+class OrionBaseAction(Action):
+    def __init__(self, config):
+        super(OrionBaseAction, self).__init__(config)
+
+        self.swis = None
```
Kami: Maybe `client` might be a better name for this variable :)

packs/orion/actions/lib/actions.py
```diff
@@ -0,0 +1,108 @@
...
+    def get_ncm_node_id(self, caption):
+        swql = "SELECT NodeID FROM Cirrus.Nodes WHERE NodeCaption=@node"
+        kargs = {'node': caption}
+        data = self.query(swql, **kargs)
+
+        if len(data['results']) == 1:
+            try:
+                return data['results'][0]['NodeID']
+            except IndexError:
+                raise IndexError("Invalid Node")
```
Kami: Probably better to throw ValueError / Exception here and below instead of IndexError.

packs/orion/actions/lib/actions.py
```diff
@@ -0,0 +1,108 @@
...
+        elif len(data['results']) >= 2:
+            raise IndexError("Muliple Nodes match '{}' NodeCaption".format(
+                caption))
+        elif len(data['results']) == 0:
+            raise IndexError("No matching NodeCaption for '{}'".format(
+                caption))
+
+    def get_ncm_transfer_results(self, transfer_id):
+        ts = {}
```
Kami: Would be good to add a docstring to this function and mention what it does (including that it performs polling).

packs/orion/actions/lib/actions.py
```diff
@@ -0,0 +1,108 @@
...
+            swql = """SELECT TransferID, Action, Status, ErrorMessage,
+            DeviceOutput FROM NCM.TransferResults
+            WHERE TransferID=@transfer_id"""
+            kargs = {'transfer_id': transfer_id}
+
+            transfer_data = self.query(swql, **kargs)
+            status = transfer_data['results'][0]['Status']
+
+            if status == 1:
+                time.sleep(10)
```
Kami: It would also be good idea to add some logging to this function.

packs/orion/actions/ncm_config_download.yaml
```diff
@@ -0,0 +1,26 @@
...
+entry_point: 'ncm_config_download.py'
+name: "ncm_config_download"
+pack: "orion"
+
+runner_type: "python-script"
+
+parameters:
+  node:
+    type: "string"
+    description: "The node to query for state in the monitoring platform."
```
Kami: Description probably needs to be updated to match this action :)

packs/orion/actions/ncm_config_download.yaml
```diff
@@ -0,0 +1,26 @@
...
+    required: true
+  platform:
+    type: "string"
+    description: "The Orion Platform to act on."
+    required: true
+  configs:
+    description: "Which config(s) to download from NCM."
+    type: "array"
+    enum:
+      - ["running"]
```
Kami: Does `enum` property works with lists?

I never tested it myself before. I image it should work with the API and CLI, but I'm not sure about WebUI.

/cc @enykeev

packs/orion/actions/status.yaml
```diff
@@ -0,0 +1,17 @@
+---
+description: "Query Solarwinds Orion for a node's status (i.e. Up/Down)"
+enabled: true
+entry_point: 'status.py'
```
Kami: Minor style / consistency thing.

We tend to prefix action names which return a single result with `get_` (e.g. `get_status` / `get_node_status`) and actions which return multiple results with `list_`.

packs/orion/aliases/ncm_download.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - display: "orion ncm config-download {{platform}} {{node}}"
+    representation:
+      - "orion ncm config-download {{platform}} {{node}}"
+ack:
+  enabled: true
+  append_url: false
+  format: "Downloading NCM config's for {{ execution.parameters.node }} you..."
+result:
+  format: |
+    {% if execution.status == 'succeeded' %}
```
Kami: Nice, I like that you included some default aliases :)

packs/orion/requirements.txt
```diff
@@ -0,0 +1 @@
+orionsdk
```
Kami: Probably a good idea to pin it to a more specific version in case the client API changes (e.g. `orionsdk>=0.0.3,<0.1`)

jjm:
We are getting there, not had as much time as I wanted and adding testing highlighted some issues :+1:.

Aiming to get to the point that we will be able to merge next week, currently we have:
- 3 aliases
- 1 rule
- 6 actions.

I've got plans for another 3 actions and 2 aliases, but they are likely to follow later so that we can complete the merge.

packs/orion/actions/lib/actions.py
```diff
@@ -0,0 +1,212 @@
...
+                  WHERE Caption=@caption"""
+        kargs = {'caption': caption}
+        caption_data = self.query(swql, **kargs)
+
+        swql = """SELECT NodeID, IPAddress FROM Orion.Nodes
+                  WHERE IPAddress=@ip_address"""
+        kargs = {'ip_address': ip_address}
+        ip_data = self.query(swql, **kargs)
+
+        if len(caption_data['results']) >= 1 or len(ip_data['results']) >= 1:
```
Kami: Would it make sense to re-arrange this a bit and return immediately if first query returns any result?

This way we could reduce total numbers of queries from 2 to 1 in case the first query already matches.

In any case, that's not a big deal, probably just a pre-mature optimization :)

packs/orion/actions/lib/actions.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns: the NodeID (int)
+        """
+        swql = "SELECT NodeID FROM Orion.Nodes WHERE Caption=@caption"
+        kargs = {'caption': caption}
+        data = self.query(swql, **kargs)
+
+        if len(data['results']) == 1:
+            try:
+                return data['results'][0]['NodeID']
+            except IndexError:
```
Kami: Might be a good idea to debug log result on failure to make it easier to debug those scenarios (e.g. is the array empty or the first item in the array is missing NodeId key...)

packs/orion/actions/lib/actions.py
Kami: This still applies :)

It's also probably a good idea to make sleep delay user-configurable via method argument and action parameter...

packs/orion/actions/ncm_config_download.yaml
Kami: Did you ever get this to work?

packs/orion/rules/node_create_webhook.yaml
```diff
@@ -0,0 +1,32 @@
+---
+name: "node_create_webhook"
+pack: "orion"
+description: "Trigger orion.node_create via a webhook"
+enabled: true
```
Kami: Probably a good idea to leave rule disabled by default.

Kami:
Thanks :+1: 

I've added some minor comments, mostly minor stuff.

packs/orion/actions/lib/actions.py
jjm: I've make it configurable in the method, but I'm not sure how useful it will be to push that up to into the action.

packs/orion/actions/ncm_config_download.yaml
jjm: It works everywhere other than the WebUI, so I'll change to a `.` seperated list and an comment to say array's & enums don't play nice.

packs/orion/rules/node_create_webhook.yaml
jjm: True, I'll just disable.

packs/orion/actions/lib/actions.py
jjm: That's a good idea, I can add some extra logging so you can see which one it was.

jjm:
Confirmed that the pack can be deployed via ChatOps with the following command:

`! pack deploy pulsant/st2contrib packs orion branch jjm/add-orion-pack`

I think we are good to merge (:+1:).

packs/orion/actions/ncm_config_download.py
```diff
@@ -27,6 +27,10 @@ def run(self, node, platform, configs):
         results = {}
         node_ids = []
 
+        # Split into an Array, as enum and array type does not play nice
+        # in the WebUI.
+        configs = configs.split(",")
```
Kami: Can we make this parameter an array without an enum attribute in the action metadata?

Kami: Other changes look good so I will go ahead and merge it into master, but I would appreciate if you can open up a new PR when this parameter type is changing.

The problem is that not using the native type in the metadata kinda breaks the whole abstraction so now user needs to know implementation detail (either check the code or description aka something which can't be done automatically) to know what type the parameter is.

jjm: There is a limited number of config names supported on the NCM side, which is why I went with an enum so it can be extended hopefully when more are supported. However not all devices will support the two supported config names. so it's not a good idea to force it to going with an immutable parameter.

jjm: I've checked and customer config types are possible now in NCM, so the enum should not be present. So I've changed back to an array and dropped the enum.

Kami:
Merged, thanks!

(I will fix a minor flake8 issue locally)

PR #468, opened 2016-04-08
Migrate RabbitMQSensor to an active sensor

manasdk:
## Pack (replace this with the pack's name)
rabbitmq 

### Status 

Replace this: describe the pack and PR status. Examples: 
- pack extension
- bugfix

### Description

* Rename to RabbitMQQueueSensor since thats what it really is and added
  some comments to that effect.
* Changed the config and made a bit more structured to accomodate for
  different sensors that support unrelated routing topologies.

### DIFF
--- a/packs/rabbitmq/sensors/queues_sensor.py
+++ b/packs/rabbitmq/sensors/queues_sensor.py
@@ -70,11 +70,13 @@ class RabbitMQQueueSensor(Sensor):
         # Setup Qs for listening
         for queue in self.queues:
             self.channel.queue_declare(queue=queue, durable=True)
-            cb = lambda ch, method, properties, body: self.callback(
-                ch, method, properties, body, queue)
-            self.channel.basic_consume(cb, queue=queue)
 
-    def callback(self, ch, method, properties, body, queue):
+            def callback(ch, method, properties, body):
+                self._dispatch_trigger(ch, method, properties, body, queue)
+
+            self.channel.basic_consume(callback, queue=queue)
+
+    def _dispatch_trigger(self, ch, method, properties, body, queue):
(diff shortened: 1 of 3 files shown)


packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -1,42 +1,62 @@
...
-        self.deserialization_method = self._config['sensor_config']['deserialization_method']
+
+        queue_sensor_config = self._config['sensor_config']['rabbitmq_queue_sensor']
+        self.queues = queue_sensor_config['queues']
+        if not isinstance(self.queues, list):
+            self.queues = [self.queues]
+        self.deserialization_method = queue_sensor_config['deserialization_method']
 
         supported_methods = DESERIALIZATION_FUNCTIONS.keys()
         if self.deserialization_method and self.deserialization_method not in supported_methods:
```
Kami: I don't see this method actually being used somewhere to de-serialize the message. Or am I missing something?

packs/rabbitmq/sensors/queues_sensor.py
```diff
@@ -1,42 +1,62 @@
...
+    def run(self):
+        self._logger.info('Starting to consume messages from RabbitMQ for %s', self.queues)
+        # run in an eventlet in-order to yield correctly
+        gt = eventlet.spawn(self.channel.start_consuming)
+        # wait else the sensor will quit
+        gt.wait()
+
+    def cleanup(self):
+        if self.conn:
+            self.conn.close()
```
Kami: We probably also want to call `self.gt.kill()` here before closing the connection, right?

Kami:
Added some comments, but overall, LGTM :+1:

PR #469, opened 2016-04-08
Icinga2 Integration pack, v0.2.0, initial commit

emptywee:
### Status 

- new pack, work in progress, not yet complete

### Description

Icinga2 Integration pack. Sensor to receive state change events, actions to get status, hosts and services information.

P.S. I am a newbie, so don't expect much from the first time :-P All contributions are welcome.


Kami:
Nice, thanks :+1: 

Let us know when it's ready for review and we will have a look.

emptywee:
Well, works for me so far, but I'd like somebody with more python knowledge to take over and refactor/rewrite/make it nicer and add more features :-) So anyone is welcome to review, modify, add and extend!

packs/icinga2/actions/lib/client.py
```diff
@@ -0,0 +1,128 @@
...
+  def abort_session(self):
+    self.keep_trying = 0
+    self.conn.close()
+
+  def setup_connection(self):
+
+    if self.conn:
+      self.conn.close()
+    self.headers = ['Accept: application/json', 'Expect:', 'Connection: close']
+    self.conn = pycurl.Curl()
```
Kami: Any particular reason why you used pycurl and not requests?

Requests is much simpler and requires a lot less boiler-plate code - http://docs.python-requests.org/en/master/

Kami: Basically, by switching to requests (unless I'm missing something) you should be able to reduce this whole file to couple of lines of code :)

packs/icinga2/sensors/lib/client.py
```diff
@@ -0,0 +1,136 @@
+import pycurl, json
+import urllib
+import signal
+import sys
+import time
+
+POST_PARAMS = {}
+
+class Client:
```
Kami: Same here and it also looks like a lot of code is duplicated. Can rid of most of it by switching to requests (mall other packs use requests so there should be plenty of usage examples around) :)

emptywee: Sure :) I just googled for the python examples of long polling and that was the first one I stumbled upon. That is the only reason I used py_curl :-)

Kami: I see :)

If you need to poll you can achieve same thing with a while loop (or similar) + some requests code.

Kami:
@emptywee How is this PR looking?

Do you plan to work on it some more? It would be great if you could address some of the comments (switching to requests, etc. would be great), but if you can't do that, the minimum would be to at least get the lint and other Travis checks stuff to pass so we can merge it into master :)

emptywee:
@Kami I am okay with whatever is better for the pack. Currently not planning to work on it, since I am getting a new job, and probably Icinga2 is not the tool they are using for monitoring. If you feel this can be switched to requests, then okay. It's just I am not sure if you will be able to test it thoroughly :-)

LindsayHill:
I'm closing this one for now, and picking it up in #552

PR #470, opened 2016-04-08
[salt] Better Positional Args Handling + Docs

jcockhren:
### Status 
bugfix

### Description

- Fixed execution module actions that failed to collect values passed to the `module` keyword.
- Improved docs with example of sending multiple positional arguments.
- Better handling of positional arguments when processing action call.

- [x] Tests (not required but really recommended)

### DIFF
--- a/packs/salt/tests/test_action_local.py
+++ b/packs/salt/tests/test_action_local.py
@@ -20,7 +20,7 @@ one_arg = {
     'module': 'this.something',
     'target': '*',
     'expr_form': 'glob',
-    'args': 'os',
+    'args': ['os'],
 }
 
 multiple_args = {
@@ -56,7 +56,9 @@ class SaltLocalActionTestCase(testtools.TestCase, BaseActionTestCase):
     def test_generic_action_one_arg(self):
         self.action.run(**one_arg)
         self.assertIn('arg', self.action.data)
+        self.assertIsInstance(self.action.data['arg'], list)
 
     def test_generic_action_multiple_args(self):
         self.action.run(**multiple_args)
         self.assertIn('arg', self.action.data)
+        self.assertIsInstance(self.action.data['arg'], list)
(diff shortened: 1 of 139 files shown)


Kami:
Thanks!

Now that we have a simple unit testing framework for packs in place, it wouldn't be such a bad idea to add some tests for the base class when you get a chance - https://docs.stackstorm.com/development/pack_testing.html :)

jcockhren:
@Kami word. Will do. jumping on that now.

Kami:
Great, thanks!

packs/salt/tests/test_action_local.py
```diff
@@ -0,0 +1,64 @@
...
+
+CONFIG_DATA = {
+    'api_url': 'https://example.com',
+    'username': 'this',
+    'password': 'that'
+}
+requests_mock.Mocker.TEST_PREFIX = 'test'
+
+
+class SaltLocalActionTestCase(testtools.TestCase, BaseActionTestCase):
```
jcockhren: @Kami I think this is a good approach for pack that use requests, any thoughts?

Kami: Yeah, looks good to me :)

Although, unless I'm missing something (please correct me if I'm wrong) it doesn't look like you are actually asserting on the HTTP requests so you could simply set `requests.requests = mock.Mock()` or similar and it should still work.

Kami:
Merged into master.

Nice work with the tests and thanks again for the contribution!

packs/salt/tests/test_action_local.py
jcockhren: @Kami yes. without `register_uri` the tests will fail given `action.run` is making a call to the the Salt API. I'm just using the `self.action.data` to very we're sending the right data.

Kami: Yeah, I think just setting it to mock should also work, but I'm fine with your current approach as well :)

This way in the future we can also assert on the URL parameters and post data if needed :)

PR #471, opened 2016-04-09
Add missing property

manasdk:
## Slack

Add in missing proeprty


Kami:
LGTM, +1

(btw, how will this affect the auto-generated code, it will get overwritten next time, right?)

manasdk:
> (btw, how will this affect the auto-generated code, it will get overwritten next time, right?)

This pack is not autogenerated. I believe https://github.com/StackStorm/st2-slack is auto-generated.

PR #474, opened 2016-04-12
HPE Insight Control Server Provisioning Integration

paul-mulvihill:
### Status 

- New Pack

### Description

Atomic actions to enable the integration of Insight Control Server Provisioning functionality with ST2. 
This enables assigning attributes and build plans to a registered server for OS installation streaming.


jjm:
I've installed this from our repo onto one of our servers successfully with `! pack deploy`.

Kami:
Thanks for the contribution!

I will review it and add in-line comments shortly.

packs/hpe-icsp/actions/icsp_buildplan_apply.py
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from lib.icsp import ICSPBaseActions
+import json
+
+
+class Apply(ICSPBaseActions):
```
Kami: Minor consistency thing - we tend to suffix Action classes with Action, so `ApplyAction` in this case.

packs/hpe-icsp/actions/icsp_buildplan_apply.py
```diff
@@ -0,0 +1,72 @@
...
+    def run(self, buildplan_uris, server_data, connection_details):
```
Kami: Since  it looks like that `connection_details` argument is optional, you should default it to none `connection_details=None`.

packs/hpe-icsp/actions/icsp_buildplan_apply.yaml
```diff
@@ -0,0 +1,28 @@
+---
+  name: "icsp_buildplan_apply"
+  runner_type: "run-python"
+  description: "Apply Buidlplan to Specified Servers"
+  enabled: true
+  entry_point: "icsp_buildplan_apply.py"
+  parameters:
+    buildplan_uris:
+      type: "array"
+      description: "Comma seperated list of build plan URIs"
```
Kami: Hm, the parameter says URIs, but looking at the code it seems like user just needs to pass a list of integers.

If that's the case, you can also use the `items` jsonschema attribute (http://spacetelescope.github.io/understanding-json-schema/reference/array.html) and it would also be good to update the parameter name (`buildplan_ids` or similar) since URIs usually also contain scheme and other information.

packs/hpe-icsp/actions/icsp_buildplan_apply.yaml
```diff
@@ -0,0 +1,28 @@
...
+      required: true
+      position: 1
```
Kami: `position` is only used for local runner script actions, but it also doesn't break anything if you leave it here (in the future we might use it for sort order or similar :)).

packs/hpe-icsp/actions/lib/icsp.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self, config):
+        super(ICSPBaseActions, self).__init__(config)
+        self.icsp_host = config['host']
+        self.icsp_user = config['user']
+        self.icsp_pass = config['pass']
+        self.icsp_apiv = config['apiv']
+        self.icsp_sslverify = config['sslverify']
+
+    def checkResults(self, results):
```
Kami: Minor style / consistency things - we usually use underscore notation for method names (`check_results`, `set_connection`, etc.).

packs/hpe-icsp/actions/icsp_buildplan_apply.py
```diff
@@ -0,0 +1,72 @@
...
+                raise ValueError("Build plans must be \
+                                 Integers (comma seperated)")
+
+            pload["osbpUris"].append("/rest/os-deployment-build-plans/%s"
+                                     % plan)
+
+        for server in server_data:
+            data = {}
+            pdata = {}
+            data['serverUri'] = "/rest/os-deployment-servers/%s" % server
```
Kami: Minor style thing - usually it's a good practice to wrap format string arguments in parenthesis (e.g. `% (servers)`), this way things won't break if you add another format string parameter, but forgot to add parenthesis.

packs/hpe-icsp/actions/icsp_buildplan_apply.py
```diff
@@ -0,0 +1,72 @@
...
+                pdata['hostName'] = server_data[server]['hostname']
+            if "domain" in server_data[server]:
+                pdata['domain'] = server_data[server]['domain']
+            if "workgroup" in server_data[server]:
+                pdata['workgroup'] = server_data[server]['workgroup']
+            data['personalityData'] = pdata
+
+            pload['serverData'].append(data)
+
+        payload = json.dumps(pload)
```
Kami: If the API always takes JSON, you could also handle serialization inside `icspPOST` method, but that's just an option and I'm also OK if you leave it as it is.

packs/hpe-icsp/actions/icsp_buildplan_apply.py
```diff
@@ -0,0 +1,72 @@
...
+        try:
+            results = self.icspPOST(endpoint, payload)
+        except Exception as e:
```
Kami: If you want to preserve a traceback from the original exception and since you aren't really changing the error message a lot, you could get rid of this except and just propagate the original exception (this way original traceback won't get lost).

packs/hpe-icsp/actions/icsp_buildplan_get.py
```diff
@@ -0,0 +1,38 @@
...
+            self.setConnection(connection_details)
+        self.getSessionID()
+
+        endpoint = "/rest/os-deployment-build-plans"
+        results = self.icspGET(endpoint)
+        plans = []
+        for plan in results["members"]:
+            if plan_names:
+                for name in plan_names:
+                    if name.lower() == str(plan["name"]).lower():
```
Kami: To make this a bit more readable and easier to test you could refactor this functionality in a separate method (e.g. `def _to_plan(self, plan_obj)`.

packs/hpe-icsp/actions/icsp_ca_get.yaml
```diff
@@ -0,0 +1,13 @@
+---
+  name: "icsp_ca_get"
+  runner_type: "run-python"
+  description: "Retreive CA Certificate from ICSP"
```
Kami: Out of curiosity since it's not really immediately clear from the code - what does this action return (an object or something else)?

packs/hpe-icsp/actions/icsp_job_status_get.yaml
```diff
@@ -0,0 +1,26 @@
...
+    job_id:
+      type: "integer"
+      description: "URI for Job, blank will return all jobs listed"
+      required: false
+      position: 0
+    monitor:
+      type: boolean
+      description: "Monitor Progress of The Job. False = One time check"
+      required: false
+      default: false
```
Kami: Usually it's also a good practice to specify default argument values in the method signature.

I know this means some duplication, but it makes the Python action a bit less coupled to StackStorm and easier to test in a standalone fashion...

packs/hpe-icsp/actions/icsp_job_status_get.py
```diff
@@ -0,0 +1,57 @@
...
+        if not job_id:
+            for job in jobs['members']:
+                jobid = job["uri"].split("/")[-1]
+                output[jobid] = job['state']
+        else:
+            status = jobs['state']
+            if monitor:
+                jobid = jobs["uri"].split("/")[-1]
+                while status == "STATUS_ACTIVE":
+                    eventlet.sleep(120)
```
Kami: Probably a good idea to make timeout user configurable aka making it an action parameter :)

packs/hpe-icsp/actions/icsp_job_status_get.py
```diff
@@ -0,0 +1,57 @@
...
+                    jobs = self.icspGET(endpoint)
+                    status = jobs['state']
+                if status == 'STATUS_SUCCESS':
+                    output[jobid] = jobs['state']
+                else:
+                    raise Exception("%s: %s" % (jobid, status))
+            else:
+                jobid = jobs["uri"].split("/")[-1]
```
Kami: Since this statement is repeated in multiple places it wouldn't be such a bad idea to refactor it in a utility method.

packs/hpe-icsp/actions/icsp_mid_get.py
```diff
@@ -0,0 +1,47 @@
...
+        servers = getresults["members"]
+        results = []
+        # Checking arrays are set otherwise errors occur later
+        # when called within flows missing arrays cause python errors
+        if not serialnumber:
+            serialnumber = []
+        if not uuid:
+            uuid = []
+        for server in servers:
+            if (server["uuid"] in uuid) or\
```
Kami: I assume there is no way to retrieve information for a single server and that's why you need to do late filtering here?

packs/hpe-icsp/actions/icsp_mid_get.py
```diff
@@ -0,0 +1,47 @@
...
+                    (server["serialNumber"] in serialnumber):
+                if (len(uuid) + len(serialnumber)) == 1:
```
Kami: I'm not sure I totally understand this line - both of those arguments appear to be array, so this if statement would be throw if only one argument is provided (either uuid or serialnumber) and it has a single item, right?

packs/hpe-icsp/actions/icsp_mid_get.yaml
```diff
@@ -0,0 +1,23 @@
+---
+  name: "icsp_mid_get"
+  runner_type: "run-python"
+  description: "Retreive MID for a given server"
```
Kami: It appears this action returns a list and allows you to retrieve information for multiple servers.

If that's the case it would be good to update the description to reflect that :)

packs/hpe-icsp/actions/icsp_server_attributes_get.py
```diff
@@ -0,0 +1,60 @@
...
+
+class GetServerAttributes(ICSPBaseActions):
+    def run(self, mids, connection_details,
+            attribute_key, attribute_type="all"):
+        if connection_details:
+            self.setConnection(connection_details)
+        self.getSessionID()
+        endpoint = "/rest/os-deployment-servers"
+        results = {}
+        for mid in mids:
```
Kami: Probably a good idea to refactor this on a utility method on a base class since it's repeated across multiple actions.

packs/hpe-icsp/actions/icsp_server_attributes_get.yaml
```diff
@@ -0,0 +1,33 @@
+---
+  name: "icsp_server_attributes_get"
+  runner_type: "run-python"
+  description: "Retrieve Attributes set against Server"
+  enabled: true
+  entry_point: "icsp_server_attributes_get.py"
+  parameters:
+    mids:
+      type: "array"
+      description: "Server MID list, Comma seperated."
```
Kami: I would personally remove "comma separated" part since this is a client (CLI in this case) implementation details.

packs/hpe-icsp/actions/lib/icsp.py
```diff
@@ -0,0 +1,98 @@
...
+        p = requests.post(url, headers=headers,
+                          json=payload, verify=self.icsp_sslverify)
+        results = p.json()
+        self.checkResults(results)
+        self.icsp_sessionid = results["sessionID"]
+        return results["sessionID"]
+
+    def icspGET(self, endpoint):
+        url = 'https://%s%s' % (self.icsp_host, endpoint)
+        headers = {'Auth': self.icsp_sessionid,
```
Kami: Minor style thing - I would personally move assignment of those base headers somewhere else (e.g. in a constructor so it's done only once) and then do something like - `p = requests.get(url, headers=self.base)headers, verify=self.icsp_sslverify)`.

packs/hpe-icsp/actions/icsp_job_status_get.yaml
Kami: Maybe `poll` would be a better name for this argument.

Kami:
Thanks.

I've added some comments in-line.

There seems to be quite a lot of filtering, etc. going on inside the actions so some tests would be great if and when you get a chance (I think the best approach would be to save API responses to `.json` files and use them as fixture inside the tests).

packs/hpe-icsp/actions/icsp_mid_get.py
paul-mulvihill: Unfortunately it's all servers or server based on MID.. so i need to pull them all in this action :(

packs/hpe-icsp/actions/lib/icsp.py
paul-mulvihill: done, will check in soon

packs/hpe-icsp/actions/icsp_buildplan_apply.yaml
paul-mulvihill: so far i've been using it to control the order or the parameter within the st2 ui.

packs/hpe-icsp/actions/icsp_buildplan_apply.yaml
paul-mulvihill: switched to id. Although not using the "items" feature yet as last time i used it there was a bug in the code where it couldn't map the elements the items type.

packs/hpe-icsp/actions/icsp_buildplan_apply.py
paul-mulvihill: done, will check in soon.

packs/hpe-icsp/actions/icsp_buildplan_apply.py
paul-mulvihill: I had updated this just not managed to check it in yet.. will do soon

packs/hpe-icsp/actions/icsp_buildplan_apply.py
paul-mulvihill: done, will check in soon

packs/hpe-icsp/actions/icsp_ca_get.yaml
paul-mulvihill: I'll admit this was just a additional action I created when I was trying to deal with CA's to enable ssl. i needed a way to pull the ICSP server cert so it's just returns the contents of the get request.

packs/hpe-icsp/actions/icsp_job_status_get.py
paul-mulvihill: done. will commit shortly

packs/hpe-icsp/actions/icsp_mid_get.py
paul-mulvihill: i do check that the arrays have been provided prior to this and set blank arrays if they are not given. meaning the len returns 0

packs/hpe-icsp/actions/icsp_job_status_get.py
paul-mulvihill: done will commit soon.

packs/hpe-icsp/actions/icsp_job_status_get.yaml
paul-mulvihill: When you mentiond default arguments which parameter do you mean?
Job id can be empty. and the default for monitor is supposed to be "false"
Unless it's a requirement in my head monitor sounds more fitting to poll. but if it's that important i can adjust.

Kami: Yeah, i'm also fine with keeping the existing name.

With default argument values I mean that every parameter which has a default value in the metadata should have a default value in the Python action file inside the run method signature.

Kami:
Thanks for addressing the feedback and the nice work overall :+1: 

There are still some minor improvements which could be made with regards to code refactoring, but this is not a blocker and can be addressed in a future when / if time permits so I will go ahead and merge it into master.

PR #475, opened 2016-04-13
Add Check_MK Pack

codyaray:
## Check_MK Pack

Check_MK is a monitoring solution in the tradition of Nagios.

### Status 

- new pack, MVP

### Description

Skeleton of a Check_MK pack. Currently only provides the minimum integration to configure Check_MK to send notification events to StackStorm. Sufficient for taking action on alerts. Does not yet support two-way communication, i.e., acking alerts.


Kami:
Great, thanks for the contribution. It looks like a good start :)

codyaray:
Any idea why this failed?

My hunch is that it tries to auto-load the (example) rules but fail because it doesn't auto-load the provided trigger? Does that sound right?

```
2016-04-13 18:19:00,167 WARNING [-] Failed to register rules: A simple trigger should have been created when registering triggertype. Cannot create trigger: {'type': 'check_mk.event_handler', 'parameters': {}, 'pack': 'check_mk'}.
  File "/tmp/st2/st2common/bin/st2-register-content", line 22, in <module>
    sys.exit(content_loader.main(sys.argv[1:]))
  File "/tmp/st2/st2common/st2common/content/bootstrap.py", line 284, in main
    register_content()
  File "/tmp/st2/st2common/st2common/content/bootstrap.py", line 250, in register_content
    register_rules()
  File "/tmp/st2/st2common/st2common/content/bootstrap.py", line 186, in register_rules
    raise e
st2common.exceptions.triggers.TriggerDoesNotExistException: A simple trigger should have been created when registering triggertype. Cannot create trigger: {'type': 'check_mk.event_handler', 'parameters': {}, 'pack': 'check_mk'}.
Registering resources for pack check_mk failed
make: *** [.packs-resource-register] Error 1
```

manasdk:
@codyaray Its because `packs/check_mk/triggers/event_handler.yaml` is not registered. In fact AFAIK the `triggers` folder is not processed.

The TriggerTypes are registered through the sensor registration or via api e.g. https://github.com/StackStorm/st2contrib/blob/master/packs/circle_ci/sensors/webhook_sensor.yaml#L5.

codyaray:
So what do I change? Anything?

Kami:
Right now you have two options:

1. Move trigger creation inside the check script (not ideal, but that's what nginx script does right now)
2. Create a dummy sensor YAML metadata file which registers this trigger. You can use some mock / dummy value for sensor name, class name, etc.

I will also discuss it with @manasdk and others today. I think we should still support registering triggers which don't belong to the any sensor (e.g. adding a YAML file to `<pack name>/triggers/` directory or similar).

I think this would be good since approach 1) breaks "infrastructure as code" approach and approach 2) is more of a hack.

codyaray:
Okay, I'll just sit tight and wait for auto-registration for triggers then. :)

codyaray:
Did ya'll figure anything out for auto-registration of triggers? Should I create an issue to track this? @Kami @manasdk

Kami:
https://github.com/StackStorm/st2/pull/2752 has been merged into trunk so we can merge this PR once all the tests pass.

It means this pack will only work with StackStorm v1.5 so it's probably a good idea to indicate this in the readme.

codyaray:
This should still be compatible with pre-1.5 but you'll have to manually register the trigger as documented in the README. This seems to be still failing at the trigger registration though. Any ideas?

Kami:
Can you gist the output please and I will have a look.

codyaray:
https://travis-ci.org/StackStorm/st2contrib/jobs/140863911 :)

Kami:
Hm, thanks.

Thanks weird, I will dig into it later.

Kami:
@codyaray I tracked it down to a bug in the trigger registration code.

I will work on a fix and make sure it's included in v1.5.1 :)

codyaray:
Whew. Thought it was just me. Thanks for being so on top of it!

-Cody

Kami:
Alright, fix is already in master so I will go ahead and merge this (tests use master).

I will let you know once v1.5.1 with fix is out (probably in two weeks or so).

PR #477, opened 2016-04-14
Fix run.py to not assume tokens is present in the args

manasdk:
### Status 

Updates to work with auto generated code.


Kami:
LGTM, :+1:

PR #478, opened 2016-04-14
rearrange data json.dumps to be performend later

manasdk:
### Status 

Fixing bug in the circle_ci.run_build action


Kami:
:+1:

PR #479, opened 2016-04-15
 parse the boto.rds.dbinstance.DBInstance object

pidah:
### Status 
- bugfix

### Description

 parse the boto.rds.dbinstance.DBInstance object so that we can retrieve useful data.


Kami:
Awesome, thanks for the contribution.

Changes look good to me so I will go ahead and merge them into master.

In the future if you get a chance it would also be great if you can add a test case for parsing (yeah, I know we don't have any existing test cases for parsing right now, but we need to start somewhere :)).

PR #483, opened 2016-04-23
Update orion.node_status to use color for alias results reflecting node stats

jjm:
### Status 

- pack extension, complete.

### Description

Use the color reflecting a node status that was already in the results and use the extra color support from 1.4 in the orion.node_status alias results.

### DIFF
--- a/packs/orion/aliases/node_status.yaml
+++ b/packs/orion/aliases/node_status.yaml
@@ -10,11 +10,13 @@ formats:
 ack:
   enabled: false
   append_url: false
-  format: "Let me check {{execution.parameters.node}} on {{execution.parameters.node}} Orion Platform for you..."
+  format: "Let me check {{execution.parameters.node}} on {{execution.parameters.platform}} Orion Platform for you..."
 result:
+  extra:
+    color: "{{execution.result.result.color|default('#439FE0')}}"
   format: |
     {% if execution.status == 'succeeded' %}
-    {{ execution.parameters.node }} is {{ execution.result.result.status }}{~}
+    {{ execution.parameters.node }} is {{ execution.result.result.status }}
     {% else %}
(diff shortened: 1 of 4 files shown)


packs/orion/aliases/node_status.yaml
```diff
@@ -10,11 +10,13 @@ formats:
...
   enabled: false
   append_url: false
-  format: "Let me check {{execution.parameters.node}} on {{execution.parameters.node}} Orion Platform for you..."
+  format: "Let me check {{execution.parameters.node}} on {{execution.parameters.platform}} Orion Platform for you..."
 result:
+  extra:
+    color: "{{execution.result.result.color|default('#439FE0')}}"
   format: |
     {% if execution.status == 'succeeded' %}
-    {{ execution.parameters.node }} is {{ execution.result.result.status }}{~}
```
emedvedev: Did you mean to remove it? On the first glance it makes more sense to have a separator here, just making sure it's not unintentional :)

jjm: I removed the separator so the whole message is in the attachment and you can see color. I'll drop a example on slack..

emedvedev:
Nice! Something else you should probably do is mention in `README.md` that this pack is only compatible with st2 1.4+. It's very new, so for some time we should keep warning people whenever an alias utilizes the new features.

jjm:
Very good suggestion, updated the README.md with something about ST2 versions.

emedvedev:
Cool! Another suggestion is to have your action give hex values for colors, since `extra.color` will hopefully be cross-adapter in the future. `good`, `danger`, etc. are Slack-only, so they won't work elsewhere.

emedvedev:
Merging, but please don't forget about the colors. :)

Issue #363, opened 2016-01-13
[ansible] Consider removing default `sudo: true`


armab:
Consider removing default `sudo: true` in Ansible pack.
Verify/Test if all possible cases/showcases still works well and all user stories are still covered.

Discussion in #community: https://stackstorm-community.slack.com/archives/community/p1452701218012208

PR #489, opened 2016-05-05
Remove sudo parameter from all the local runner actions

Kami:
This pull request sudo parameter from all the local runner actions to account for this runner parameter removal in https://github.com/StackStorm/st2/pull/2671.

Note: those changes are fully backward compatible so we can merge them before we merge https://github.com/StackStorm/st2/pull/2671.

### DIFF
--- a/packs/sensu/actions/aggregate_check.yaml
+++ b/packs/sensu/actions/aggregate_check.yaml
@@ -12,7 +12,5 @@ parameters:
   age:
     type: string
     description: The number of seconds old an aggregate must be to be listed.
-  sudo:
-    immutable: true
   kwarg_op:
     immutable: true
--- a/packs/sensu/actions/aggregate_check_delete.yaml
+++ b/packs/sensu/actions/aggregate_check_delete.yaml
@@ -14,7 +14,5 @@ parameters:
     description: ''
     default: 'yes'
     immutable: true
-  sudo:
-    immutable: true
   kwarg_op:
     immutable: true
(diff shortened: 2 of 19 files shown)


Kami:
Looks like packagecloud actions use sudo and default it to true. Is this required aka things will break if I remove it?

/cc @DoriftoShoes

Kami:
Looks like there are two blockers right now - ansible and packagecloud actions.

For packagecloud I think we can probably just get rid of it, but for ansible we need to verify that it still works without sudo (I would imagine, this could cause issues since it might not be able to access inventory file, etc.).

lakshmi-kannan:
lgtm

lakshmi-kannan:
You kind have a nervous +1 from @armab to deprecate sudo in ansible pack. Internal slack link: https://stackstorm.slack.com/archives/standup/p1462455873001277 (No public access)

DoriftoShoes:
I don't remember why packagecloud uses it.  We just need to make sure the system user has the packagecloud cli set up...then it should work fine.  I am +1 to this change.

armab:
Related to: https://github.com/StackStorm/st2contrib/issues/363

Kami:
@DoriftoShoes Thanks for confirming.

As far as ansible pack goes - I will look into it (I will try to test it) and I will also update the readme and changelog and clarify this change and potential impact.

Kami:
Closing for now.

Sadly this requires more work.

PR #490, opened 2016-05-05
VSphere action de-dupication

paul-mulvihill:
### Status 
- pack extension/cleanup

### Description

Streamline of vSphere Actions, removal of action duplications,
None "Atomic" actions have been moved to stasis in order to be reviewed and atomised at a later date.
Addition of moid retrieval action.


manasdk:
I am perfectly fine with the changes you are making. The only concern I have is about backwards compatability but that is something it seems must be broken to make forward progress.

If there are complaints on why an action was removed etc then we can go down the route of "forking" this pack and calling the forked pack deprecated and focus future development efforts on this pack. I can take up the task of forking for after since I am not sure how many user of the vsphere pack are out there anyway.

paul-mulvihill:
Unfortunately to clean this up there is going to be that risk.. I haven't removed anything that isn't covered in one of the existing actions. so hopefully it won't be a problem. Anything that is less "atomic" has been put in stasis so not lost. I didn't see any comments when i asked in the comunity how many are using this pack yet. but it could have got lost easily in the history. so no idea how much it is used.

manasdk:
Lets go for it - Let me know when you feel you are ready to merge this PR. We shall keep trying to make forward progress.

paul-mulvihill:
lets do it then. I've not made any recent changes and can't think of any others to do before it's merged.

PR #492, opened 2016-05-06
Add Cloudflare pack

jjm:
### Status 

- new pack, complete

### Description

Add a pack for Cloudflare (https://www.cloudflare.com/). Currently covers a single API call, but the API has many more and they will be added over time.


emedvedev:
Oh nice! You wanna add some more actions/aliases before I merge, or would you rather do it later in a separate PR? I don't mind merging, since there's `lib` and tests and everything (awesome job on it, by the way).

jjm:
Thanks, pushing the bonds with tests first time I've used `requests_mock`!

Assuming linting is fine (I hope they are as I ran it locally), please go ahead and merge. As currently I'm not sure how long before I can add more actions, so I'll do those in a separate PR.

armab:
Cloudflare! :star2:

jjm:
Thanks!

emedvedev:
:heart:

Kami:
Thanks for the contribution.

In the future, if you are ever interested in the DNS / domain related functionality of CloudFlare, you can also look at the CloudFlare Libcloud DNS driver - https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/cloudflare.py

packs/cloudflare/tests/test_action_get_ips.py
```diff
@@ -0,0 +1,128 @@
...
+api_host: "api.cloudflare.com"
+"""
+
+MOCK_CONFIG_FULL = """
+api_host: "mock://api.cloudflare.com"
+"""
+
+MOCK_DATA_INVALID_JSON = "{'dd': doo}"
+
+MOCK_DATA_SUCCESS = """
```
Kami: Minor thing - I usually prefer to put raw response fixtures like that in a file (e.g. `packs/cloudflare/tests/fixture/list_ips_success.json` or similar) and load file content in the code.

This way it's a bit easier to update those files and make sure they indeed contain valid JSON (e.g. we can add a CI step which runs syntax check on all the *.json files or similar).

If needed, we can also add some helper functions to the base test class which allows you to easily load content from a file (although, even without that function it's only 1-3 lines of Python, but the function could at least enforce a convention for fixture file locations - e.g. `packs/<pack name>/tests/fixtures/<blah>/<fixture file name>`).

jjm:
Thanks that will save us a chunk of work. I'll try to make sure this pack does not overlap with libcloud DNS features. Or may be even do PRs to libcloud if that makes sense.

packs/cloudflare/tests/test_action_get_ips.py
jjm: Helper functions for that sort of thing would be useful, this is my first attempt at fixtures for requests, so anything to make it neater would be cool.

Preventing repeat of a few lines of code and enforcing location sounds great. If / when you have that in master I'll happily update the packs I've worked on.

Kami: Will let you know :)

jjm: Thanks, 

I've followed the suggestion and raised #493 to move the fixtures into separate files.

PR #496, opened 2016-05-17
hpe-icsp refinement

paul-mulvihill:
### Status 
- pack extension

### Description

Refining Pack post real world application. Reduced load on mistral for multi server buildplan application.
Centralise get_mid function and adapt mid_get and server data format to use it.
Expanding server data format action to accept different Identifier types

Previous version of "ids to os" flow used "with item" feature to handle multiple servers. This method could hog Mistral delaying other calls. Update uses batch features of ICSP allowing this iterative work moved to actions.

Main action interface changes
* icsp_mid_get: merged serialnumber and uuid fields
* icsp_server_data_format: change mid field to identifiers and added identifier type

### DIFF
--- a/packs/hpe-icsp/actions/icsp_server_data_format.py
+++ b/packs/hpe-icsp/actions/icsp_server_data_format.py
@@ -17,7 +17,7 @@ from lib.icsp import ICSPBaseActions
 
 
 class FormatServerData(ICSPBaseActions):
-    def run(self, identifiers, identifier_type,  hostnames,
+    def run(self, identifiers, identifier_type, hostnames,
             domains=None, workgroups=None, connection_details=None):
         if identifier_type == "mid":
             for n in identifiers:
--- a/packs/hpe-icsp/actions/icsp_server_data_format.py
+++ b/packs/hpe-icsp/actions/icsp_server_data_format.py
@@ -17,11 +17,12 @@ from lib.icsp import ICSPBaseActions
 
 
 class FormatServerData(ICSPBaseActions):
-    def run(self, identifiers, identifier_type,  hostnames, domains=None, workgroups=None, connection_details=None):
+    def run(self, identifiers, identifier_type,  hostnames,
+            domains=None, workgroups=None, connection_details=None):
         if identifier_type == "mid":
             for n in identifiers:
                 try:
-                  test = int(n) 
+                    int(n)
                 except:
                     raise ValueError("Identifier provides is not a MID")
             mids = identifiers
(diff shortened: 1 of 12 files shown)


packs/hpe-icsp/actions/icsp_server_data_format.py
```diff
@@ -17,7 +17,19 @@
 
 
 class FormatServerData(ICSPBaseActions):
-    def run(self, mids, hostnames, domains=None, workgroups=None):
+    def run(self, identifiers, identifier_type, hostnames,
+            domains=None, workgroups=None, connection_details=None):
+        if identifier_type == "mid":
+            for n in identifiers:
+                try:
```
Kami: Would be nice to refactor this in a separate utility validation function, but it's not such a big deal.

packs/hpe-icsp/actions/lib/icsp.py
```diff
@@ -63,14 +63,31 @@ def get_sessionid(self):
...
 
+    def get_mids(self, ids, idtype):
+        endpoint = "/rest/os-deployment-servers"
+        getresults = self.icsp_get(endpoint)
+        servers = getresults["members"]
+        mids = []
+        for id in ids:
+            for server in servers:
+                if idtype == 'serialnumber':
+                    if ((server["serialNumber"] == id) and
```
Kami: This double loop and nesting is a bit hard to read and probably also not that efficient...

I think you could probably get rid of that outter loop and do something like:

```python
for server in servers:
    if (idtype == 'serialnumber' and server["serialNumber"] in ids) and (server["mid"] not in mids):
        mids.append(int(server["mid"]))
    elif (idtype == 'uuid' and server["uuid"] in ids) and (server["mid"] not in mids):
        mids.append(int(server["mid"]))
```

Kami:
Thanks.

I've added some in-line comments.

packs/hpe-icsp/actions/lib/icsp.py
paul-mulvihill: i did used to have just one for loop (server in servers) although because of the way this function is called i need to return mid values in the same order as the ids they relate to in the parameter. using server in servers means that the mids are passed back based on the order the icsp server returns them not what the person inputs. Will see if i can think of a better cleaner way.

Kami: Ah, ok, thanks for clarifying it.

Probably also a good idea to add the same comment to the code - aka why it's done the way it is :)

paul-mulvihill: refactored the if statement and added a comment as to why the loops are how they are.

Kami:
Merged, thanks!

PR #497, opened 2016-05-19
s/_sensor_service/sensor_service

lakshmi-kannan:


Kami:
We can't merge that until everyone upgrades to the latest version of StackStorm where this change is available.

If we merged it, it would break for everyone who is not using latest version of StackStorm (that's also the limitation because we don't have pack versioning).

Sadly we are stuck with the old approach for the foreseeable future.

codyaray:
Could you keep the inconsistent way as an alias to the correct one but mark it as deprecated and to be removed at the next major upgrade? In this case, updating all the packs in st2contrib would still be a good idea. :)

Kami:
@codyaray Yes, that's what we are doing in the st2 PR - we will still leave the old alias there for the unforeseeable future :)

lakshmi-kannan:
Yeah, I woke up this morning thinking the same. I'll close this PR.

PR #498, opened 2016-06-03
Add networking_utils pack

jjm:
### Status 

- new pack, complete

### Description

Add a new pack containing some useful networking valuation actions for use in workflows. 

- [X] Installs via GitHub with packs deploy.


Kami:
Thanks!

For consistency with other pack names, please use underscore instead of - in the directory name (`-` won't work in all the places since we actually use directory name for pack reference right now in some places and not the actual `name` attribute from `pack.yaml`)

packs/networking-utils/actions/is_valid_asa_nameif.py
```diff
@@ -0,0 +1,38 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from st2actions.runners.pythonrunner import Action
+
+
+class IsValidNameifAction(Action):
+    def run(self, nameif):
+        """
+        Checks if nameif is valid for use on a Cisco ASA as an interface name.
```
Kami: I'm fine with this action living in this pack, but I would imagine some people would probably prefer it to be in a separate `cisco` pack :P

jjm:
That's it renamed, as requested.

packs/networking-utils/actions/is_valid_asa_nameif.py
jjm: True, it maybe it should be in a Cisco pack. However seems to fit within the validation idea behind this pack.

If others object I'm happy to remove for now.

jjm:
I've confirmed manually they all load with the right pack name and work within ST2 ok.

Kami:
Merged, thanks!

If people are not happy with cisco util action being in this pack, we can always move it later. I will also add an icon in a separate commit.

manasdk:
Awesome!

PR #499, opened 2016-06-06
Windows RPC and WinRM support for remote commands

tonybaloney:
### Status 

- pack extension, complete

### Description

Adds Winexe (RPC) remote command execution (lowest common denominator) support to the windows pack, and WinRM support. Added 2 handy commands to enable WinRM and then disable WinRM.
Both are totally bad-practice according to Microsoft, but Microsoft don't enable WinRM connection from an 'untrusted' computer, i.e. a Windows one in the same domain.


packs/windows/actions/winrm_cmd.py
```diff
@@ -0,0 +1,33 @@
...
+            server_cert_validation='ignore')
+        shell_id = p.open_shell()
+        std_out_logs = []
+        std_err_logs = []
+
+        # run the command
+        command_id = p.run_command(shell_id, command, params)
+        std_out, std_err, status_code = p.get_command_output(shell_id,
+                                                             command_id)
+        std_out_logs.append(std_out)
```
lakshmi-kannan: Not sure why this appending is needed if there is no looping.

packs/windows/actions/winrm_cmd.py
```diff
@@ -0,0 +1,33 @@
...
+        std_err_logs.append(std_err)
+        p.cleanup_command(shell_id, command_id)
+
+        p.close_shell(shell_id)
+        return std_out_logs, std_err_logs
```
lakshmi-kannan: Probably better to return structured data. I wonder how this can be consumed in action chain if there is no JSON structure.

lakshmi-kannan:
Aside from the minor comment and travis being unhappy, 👍

PR #511, opened 2016-06-28
Add ZooKeeper pack with non_blocking_lease action

codyaray:
### Status 

- new pack, minimally complete

### Description

Pack which allows integration with Apache ZooKeeper.


Kami:
Nice, thanks. :+1: 

Will review it as soon as possible.

packs/zookeeper/actions/non_blocking_lease.py
```diff
@@ -0,0 +1,49 @@
+from datetime import timedelta
+from kazoo.client import KazooClient
+from socket import gethostname
+from st2actions.runners.pythonrunner import Action
+import sys
```
lakshmi-kannan: Nit: Heh. Please order imports and group them based on PEP guidelines.

packs/zookeeper/config.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
lakshmi-kannan: Please switch to using new config.schema approach.

packs/zookeeper/config.yaml
```diff
@@ -0,0 +1,3 @@
...
+zookeeper_hosts: "localhost:2181"
+zookeeper_root: "/"
```
lakshmi-kannan: Wondering why you decided to put this in config as opposed to action param?

lakshmi-kannan:
Nice!

packs/zookeeper/config.yaml
codyaray: This was a PR before the new config.schema approach was released. ;)

packs/zookeeper/config.yaml
codyaray: We could make it overridable in action params, but this was the simplest thing that solved the immediate need. My basic reasoning is that StackStorm users don't (usually) care about URLs, so specify them system-wide. Of course, that assumption fails when you have multiple Zookeeper clusters that you want to integrate with; but we can cross that bridge when we come to it. :)

lakshmi-kannan: Yeah, I'd actually prefer to namespace it like /stackstorm/leases/ but I think anyone familiar with ZK can configure it the way they want.

packs/zookeeper/config.yaml
lakshmi-kannan: I figured. It's a suggestion :P.

packs/zookeeper/actions/non_blocking_lease.py
codyaray: Fixed. Assuming you want alphanumeric ordering within groups, which PEP-8 doesn't actually specify. https://www.python.org/dev/peps/pep-0008/#imports :)

lakshmi-kannan: yeah, the latest commit looks good.

packs/zookeeper/config.yaml
codyaray: Yeah, I set it to "/stormcron" in my config too. I could see needing to override the root path on an action-by-action basis before needing to override the host. But that can be done as a follow-up. :)

PR #512, opened 2016-06-29
Purge openstack pack and point to https://github.com/StackStorm/opens…

lakshmi-kannan:
The openstack pack in st2contrib doesn't seem to work and people are getting confused with another pack at https://github.com/StackStorm/openstack. So I decided to purge the one in st2contrib.


Kami:
Updated check scripts to skip stub packs, will go ahead and merge it into master after the tests pass.

PR #514, opened 2016-06-30
Add Elasticsearch Sensor

DoriftoShoes:
### Description

This PR adds an Elasticsearch sensor that detects if a given query returns hits that exceed a given threshold.

### DIFF
--- a/packs/elasticsearch/sensors/count_sensor.py
+++ b/packs/elasticsearch/sensors/count_sensor.py
@@ -1,0 +1,68 @@
+from st2reactor.sensor.base import PollingSensor
+from elasticsearch import Elasticsearch
+import json
+
+class ElasticsearchCountSensor(PollingSensor):
+    """
+    * self.sensor_service
+        - provides utilities like
+            get_logger() for writing to logs.
+            dispatch() for dispatching triggers into the system.
+    * self._config
+        - contains configuration that was specified as
+          config.yaml in the pack.
+    * self._poll_interval
+        - indicates the interval between two successive poll() calls.
+    """
+
+    def setup(self):
+        self.host = self._config.get('host', None)
+        self.port = self._config.get('port', None)
+        self.query_window = "%is" % self._config.get('query_window', None)
+        self.query_string = self._config.get('query_string', '{}')
+        self.count_threshold = self._config.get('count_threshold', 1)
+        self.index = self._config.get('index', '_all')
+        self._trigger_ref="elasticsearch.count_event"       
+        self.LOG = self.sensor_service.get_logger(__name__)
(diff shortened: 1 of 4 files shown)


packs/elasticsearch/sensors/count_sensor.py
```diff
@@ -0,0 +1,69 @@
+from st2reactor.sensor.base import PollingSensor
+from elasticsearch import Elasticsearch
+import json
+import time
+
+
+class ElasticsearchCountSensor(PollingSensor):
+
+    def setup(self):
+        self.host = self._config.get('host', None)
```
lakshmi-kannan: I think you can simply use self.config() as of 1.3 onwards. self._config is deprecated.

packs/elasticsearch/sensors/count_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+        self.LOG = self.sensor_service.get_logger(__name__)
+        self.query = json.loads(self.query_string)
+        self.es = None
+        
+        try:
+            self.es = Elasticsearch([{'host': self.host, 'port': self.port}])
+        except:
+            self.LOG.exception("Could not connect to elasticsearch. %s:%i" %
+                               (self.host, self.port))
+            raise Exception("Could not connect to elasticsearch. %s:%i" %
```
lakshmi-kannan: Just do raise. You'll see the original exception which is probably better.

packs/elasticsearch/sensors/count_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+        if hits.get('total', 0) > self.count_threshold:
+            payload = dict()
+            payload['results'] = hits
+            payload['results']['query'] = query_payload
+            self.LOG.info("Dispatching trigger")
+            self.sensor_service.dispatch(trigger=self._trigger_ref,
+                                         payload=payload)
+            cooldown = (self.query_window * self.cooldown_multiplier)
+            self.LOG.info("Cooling down for %i seconds" % cooldown)
+            time.sleep(cooldown)
```
lakshmi-kannan: You may want to use eventlet.sleep so you can use the cooperative yield. Otherwise, this might just block. 

Ideally, you'd add the triggers to a queue. You then have a timer thread that then scans the queue and emits the trigger. In either case, if it's in memory you'd lose events. We might have to stub this out in base class at some point so sensor authors don't have to worry about this.

cc: @Kami

Kami: IIRC, we do eventlet monkey patch in wrapper so it's fine, but yeah using `eventlet.sleep` is preferred to make it more explicit and less "magic".

PR #518, opened 2016-07-04
Fix cloudformation related parsing bug in the AWS pack

Kami:
It looks like that in newer versions of boto, you can't access `cloudformation` pack via the `boto` module anymore (might also be a bug in an upstream boto package, but didn't have time to dig in yet).

```python
Type "help", "copyright", "credits" or "license" for more information.
>>> import boto
>>> boto.cloudformation.stack.Stack
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'cloudformation'
>>> from boto import cloudformation
>>> cloudformation.stack.Stack
<class 'boto.cloudformation.stack.Stack'>
```

### DIFF
--- a/packs/aws/actions/lib/ec2parsers.py
+++ b/packs/aws/actions/lib/ec2parsers.py
@@ -1,6 +1,8 @@
 import boto
 import six
 
+from boto import cloudformation
+
 
 class FieldLists():
     ADDRESS = [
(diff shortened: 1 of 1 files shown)


estee-tew:
https://stackstorm.slack.com/archives/community/p1467637808003274

PR #519, opened 2016-07-04
Add config schema to aws pack, make it work with new improved configs

Kami:
This pull request adds config schema to the aws pack and makes it work with the new dynamic configs.

New dynamic config items don't support nested configs so I needed to make some changes the code so it's backward compatible and supports both approaches.

Hopefully in the near future we can update all the configs and reduce the nested items (keep them flat).

Here is an example of a new  `aws.yaml` config file which uses user-specific credentials, but the same shared region:

```yaml
---
aws_access_key_id: "{{user.aws_access_key_id}}"
aws_secret_access_key: "{{user.aws_secret_access_key}}"
region: "us-east-1"
```

### DIFF
--- a/packs/aws/actions/lib/action.py
+++ b/packs/aws/actions/lib/action.py
@@ -15,27 +15,47 @@ class BaseAction(Action):
     def __init__(self, config):
         super(BaseAction, self).__init__(config)
 
+        self.credentials = {
+            'region': None,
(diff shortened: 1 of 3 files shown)


estee-tew:
https://stackstorm.slack.com/archives/community/p1467721438003378

PR #524, opened 2016-07-22
vsphere Improved connect error check and item get function

paul-mulvihill:
### Status 

- pack extension

### Description

Addition of new action to return list of items from vsphere instance.
Improved Exception handling and reporting on Establish Connection function. 
Extended vm_hw_basic_build workflow to take datastore flag as well as boolean value for poweron option

### DIFF
--- a/packs/vsphere/tests/vsphere_base_action_test_case.py
+++ b/packs/vsphere/tests/vsphere_base_action_test_case.py
@@ -14,7 +14,7 @@
 
 import yaml
 import json
-# from mock import MagicMock
+from mock import MagicMock
 
 from st2tests.base import BaseActionTestCase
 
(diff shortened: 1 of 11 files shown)


packs/vsphere/tests/vsphere_base_action_test_case.py
```diff
@@ -57,3 +66,11 @@ def test_run_config_old(self):
     def test_run_config_new(self):
         action = self.get_action_instance(self.new_config)
         self.assertIsInstance(action, self.action_cls)
+
+    def test_run_config_old_partial(self):
+        self.assertRaises(KeyError, self.action_cls, self.old_config_partial)
```
Kami: You can also use `assertRaisesRegexp` to also assert that a correct error message is thrown.

Kami: This is not a blocker and can be done as a future improvement :)

Kami:
Merged, thanks!

PR #529, opened 2016-08-01
[WIP] Adding MetaWeather support.

bigmstone:
### Status

new pack, work in progress
### Description

Provides programmatic access to weather data without needing API key. This will eventually replace the Google pack in ST2 chatops tutorial.
- [x] Metadata: pack.yaml, icon, structure
- [ ] Code linting
- [ ] Tests
- [ ] "Pretty print" action for clean chatops results
- [ ] Chatops Alias


packs/metaweather/actions/get_location_results.py
```diff
@@ -0,0 +1,18 @@
...
+    'GetLocationResultsAction'
+]
+
+
+class GetLocationResultsAction(Action):
+    def run(self, query):
+        results = requests.get('https://www.metaweather.com/api/location/search/?query=%s' % (query))
+
+        if results.status_code != 200:
+            raise Exception("Call to MetaWeather failed.")
```
Kami: Should also include body (or at least log it) in this case otherwise it will be very hard to debug it.

packs/metaweather/README.md
```diff
@@ -0,0 +1,8 @@
+# MetaWeather Integration Pack
+
+Pack providing integration with MetaWeather API.
+
+## Actions
+
+* ``get_location_results`` - Retrieve a list of locations given a search query. Ex. ZIP code, Country/City/State, etc.
```
Kami: If possible, since both actions return a dictionary, it would be great if you can include an example of the result object in the readme.

Kami: Besides that, LGTM.

What else needs to be done to take this across the finish line?

PR #532, opened 2016-08-03
Change params type from list to array

LindsayHill:
### Status 

- bugfix

### Description

Bug fix  - incorrect parameter type

- [ X] Version bump (required for changed packs)

### DIFF
--- a/packs/windows/actions/winrm_cmd.yaml
+++ b/packs/windows/actions/winrm_cmd.yaml
@@ -22,8 +22,8 @@
       description: "The command to run"
       required: true
     params:
-      type: "list"
-      description: "list of parameters for the command"
+      type: "array"
+      description: "array of parameters for the command"
       required: true
     port:
       type: "integer"
--- a/packs/windows/pack.yaml
+++ b/packs/windows/pack.yaml
@@ -6,6 +6,6 @@ keywords:
   - wmi
   - windows management interface
   - wql
-version : 0.2
+version : 0.3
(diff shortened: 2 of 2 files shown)


lakshmi-kannan:
Good catch!

Kami:
Good catch :+1: 

Not sure how the tests didn't originally catch it though...

LindsayHill:
@Kami I've been wondering about the tests too. Even if we don't pick it up when the tests run, I think it should be picked up when you attempt to register that action. From what I could tell, it lets you register the action, but doesn't fail until you try to use it.

Kami:
Ah, that's weird, indeed - registering a resource should indeed fail on invalid schema (invalid parameter type as in this case).

In any case, it's possible it's some edge case we don't handle correctly, will dig in at some point.

LindsayHill:
@Kami I just did a quick re-check with 1.6.0-4 (my original tests were with 1.5). Looks like it still allows registration of an invalid parameter type. 
I _suspect_ (but haven't tested) that if I used the new config schema it would fail.

Kami:
I was able to replicate the problem, I will look into a possible fix.

Kami:
@LindsayHill This was good catch, it looks like we never actually validated `type` attribute for action parameters correctly.

I have a fix in place and will add test to make sure this doesn't happen again in the future.

LindsayHill:
Excellent. I'm sure the fix will help someone in future

PR #538, opened 2016-08-12
Change AND to OR in the Twitter sensor

emedvedev:
Does exactly what it says.

### DIFF
--- a/packs/twitter/README.md
+++ b/packs/twitter/README.md
@@ -12,6 +12,8 @@ Pack which allows integration with Twitter.
 * ``access_token`` - Twitter API access token.
 * ``access_token_secret`` - Twitter API access token secret.
 * ``query`` - A query to search the twitter timeline for. Must be an array.
+  By default OR is applied for array items, and you can refine your results
+  further via rule criteria.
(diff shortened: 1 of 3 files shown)


Kami:
This works for me, although it's also possible to achieve the same behavior using OR directly in the keyword.

For example:

```yaml
- StackStorm OR Stack_Storm
```

Kami:
And no matter which approach we decide to go with, it would be good to document this with some examples in the readme :)

emedvedev:
> This works for me, although it's also possible to achieve the same behavior using OR directly in the keyword.

Surprisingly, it's not. I looked at the code, and I thought it should be like that, too, but that didn't work.

Kami:
@emedvedev I should have been more clear, should have said "it's possible to achieve the same behavior if you use a single search operator" :)

In fact, that's how the pack worked at the beginning - it only took a string, then at some point we updated it to take a list / array. But yeah, when we updated it to take a list it would make more sense to also change it to OR behavior by default, agreed :)

Kami:
:+1:

emedvedev:
Updated the readme and default config, too.

PR #543, opened 2016-08-16
Brocade VDX

bigmstone:
### Status 
new pack, finished

### Description
Brocade VDX pack for interacting with Brocade NOS devices.

### Checklist
- [x] Metadata: pack.yaml, icon, structure
- [x] Version bump
- [x] Code linting
- [x] Tests


dzimine:
@bigmstone can you please add any auto-generation tools you use into ./etc here, so that they come together. Like we do in https://github.com/StackStorm/openstack pack.

Kami:
Some ideas for future improvements for the action generator to make the code feel a bit more Pythonic and less auto-generated:

* Use camel case for class names
* Explicitly declare arguments instead of using args

In any case, not a blocker, just something we can improve on in the future.

dzimine:
good by my YOLO look. +1

Kami:
If it works, let's go ahead and merge it :+1: 

We can improve it in the future.

PR #544, opened 2016-08-17
Created sample email rule

elitegoodguy:
### Status 

Replace this: email, additional documentation as a sample rule

### Description

Sample Rule for email pack

### DIFF
--- a/packs/email/rules/sample.emailsample.yaml
+++ b/packs/email/rules/sample.emailsample.yaml
@@ -1,0 +1,25 @@
+---
+uid: 'rule:email:Sample'
+tags: []
+type:
+  ref: standard
+  parameters:
+enabled: true
+name: Sample
+trigger:
+  ref: email.imap.message
+  type: email.imap.message
+  parameters:
+criteria:
+  trigger.from:
+    pattern: johndoe@gmail.com
+    type: contains
+action:
+  ref: core.sendmail
+  parameters:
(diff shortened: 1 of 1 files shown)


elitegoodguy:
Take it easy on me first github PR

packs/email/rules/sample.emailsample.yaml
```diff
@@ -0,0 +1,25 @@
+---
+uid: 'rule:email:Sample'
```
Kami: uid is actually optional (it's automatically inferred from other fields), you can remove it from the resource metadata file.

Kami: Besides that, LGTM :+1:

packs/email/rules/sample.emailsample.yaml
```diff
@@ -0,0 +1,25 @@
...
+tags: []
+type:
+  ref: standard
+  parameters:
+enabled: false
+name: Sample
+trigger:
+  ref: email.imap.message
```
Kami: Same here, only `type` should be required

Kami:
I removed the attributes which are not needed when loading rule from disk (I assume you created and dumped it using an API), added description and sorted the attributes so they follow more standard order and merged PR into master - 667433b6f57587b83eeb3dae6114e8e103ba4964.

Thanks again!

PR #527, opened 2016-07-28
upload wechat notification plugin

Martindlut:
### Status 

- new pack, complete

### Description

Add new action which sends messages to Wechat Enterprise Account

### Checklist (tick everything that applies)

- [x] Metadata: pack.yaml, icon, structure (required for new packs)
- [x] Version bump (required for changed packs)
- [x] Code linting (required, can be done after the PR checks)
- [ ] [Tests](https://docs.stackstorm.com/development/pack_testing.html) (not required but really recommended)

PR #554, opened 2016-09-09
[WIP] Wechat

LindsayHill:
### Status
- new pack, complete
### Description

Add new action which sends messages to Wechat Enterprise Account

(based upon original PR #527 from @Martindlut


LindsayHill:
Clearly haven't fixed all linting errors yet then...

LindsayHill:
OK, at least I've figured out why `make lint` on my Mac didn't pick up all the linting errors. Problem stems from use of `printf` with `find`. https://github.com/StackStorm/st2sdk/blob/master/scripts/st2-check-pylint-pack#L93

BSD `find` doesn't allow use of `printf`. Spits out error message when `st2-check-pylint-pack` runs, but it doesn't exit with error.

We do allow for use of `greadlink` in that file, so I could update the script to use `gfind`, and tell users to install coreutils & findutils before using it on OS X.

LindsayHill:
Rather than fix all the old code, I've stripped it back to the minimum modules required for the `send_wechat` action

LindsayHill:
OK, tests passing, progress at last. Of course, it needs some testing...and some basic changes like moving auth info to config options.

PR #555, opened 2016-09-12
Correct a reference of `Travis` to `Circle` in Circle CI pack.

jjm:
### Status 

- bugfix.
- Complete.

### Description

Correct a typo in the README (so not increasing the pack version number).

Correct a typo in the Circle CI README.md.

### DIFF
--- a/packs/circle_ci/README.md
+++ b/packs/circle_ci/README.md
@@ -1,7 +1,7 @@
 # Circle CI Integration Pack
 
 Pack for integration of Circle CI into StackStorm. The pack includes the
-functionality to perform actions on Travis CI through StackStorm.
+functionality to perform actions on Circle CI through StackStorm.
 
 ## Actions
 


jvrplmlmn:
👍

jjm:
Oh dear, I'll see if I can fix that unit test as well.

Kami:
Good catch, thanks.

Kami:
Looks like the error is related to new version of requests, we encountered the same issue in StackStorm.

With new version, every header value sent via `requests` now needs to be a string. I will merge this directly into master and make the headers change myself (there are probably other places which need to be fixed / changed as well).

Kami:
There we go - 80f05c70a8039c5b613146dcb0c05468d9dcbfb0.

It's actually quite a nasty change in requests, I would imagine it affects a lot of the existing code and it's likely new version of requests will change to the old behavior in the future.

jjm:
Thanks @Kami

PR #559, opened 2016-09-16
Updated Nagios rules

vishnu81:
### Status 
-updated rules
### Description

Update the nagios rules with syntax modifications

### DIFF
--- a/packs/nagios/rules/nagios_service_load.yaml
+++ b/packs/nagios/rules/nagios_service_load.yaml
@@ -13,7 +13,6 @@ criteria:
     pattern: HARD
     type: matchregex
 action:
-  ref: check_loadavg
+  ref: linux.check_loadavg
   parameters:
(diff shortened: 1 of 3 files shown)


packs/nagios/rules/nagios_service_load.yaml
```diff
@@ -13,7 +13,6 @@ criteria:
     pattern: HARD
     type: matchregex
 action:
-  ref: check_loadavg
+  ref: linux.check_loadavg
```
Kami: Good catch, thanks.

Kami:
Merged, thanks!

Kami:
I also pushed a change so rules are disabled by default (consistent with other rules) and so we use a sample email instead of Patrick's one.

d3bbb2c8b39353d305f3eb10dbc6fce90e17aabd

PR #561, opened 2016-09-20
Vsphere st2 v2 action updates

paul-mulvihill:
### Status 
- bugfix

### Description

Pack updates to factor in ST2 V1.5-2 changes.
Some code changes impact how actions return values. Pack updated to allow functionality within st2 2.0

### DIFF
--- a/packs/vsphere/actions/vm_env_items_get.py
+++ b/packs/vsphere/actions/vm_env_items_get.py
@@ -50,12 +50,12 @@ class GetItems(BaseAction):
         itemlist = inventory.get_managed_entities(self.si_content, objecttype)
         for item in itemlist.view:
             values = {}
-            values["ID"] = item
+            values["ID"] = str(item)
             if parents:
-                values["Parent"] = item.parent
+                values["Parent"] = str(item.parent)
 
             if itemtype != 'DataCenter' and summary:
-                values["summary"] = item.summary
+                values["summary"] = str(item.summary)
 
             results[item.name] = values
 
--- a/packs/vsphere/actions/workflows/vm_hw_basic_build.yaml
+++ b/packs/vsphere/actions/workflows/vm_hw_basic_build.yaml
@@ -25,6 +25,7 @@ vsphere.vm_hw_basic_build:
     barebones_build:
       # [189, 97]
       action: vsphere.vm_hw_barebones_create
+      wait-after: 12
       input:
         vm_name: <% $.vm_name %>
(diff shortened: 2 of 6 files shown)


packs/vsphere/actions/vm_env_items_get.py
```diff
@@ -50,12 +50,12 @@ def run(self, itemtype, parents=False, summary=False, vsphere=None):
...
         for item in itemlist.view:
             values = {}
-            values["ID"] = item
+            values["ID"] = str(item)
             if parents:
-                values["Parent"] = item.parent
+                values["Parent"] = str(item.parent)
 
             if itemtype != 'DataCenter' and summary:
-                values["summary"] = item.summary
```
Kami: Hm, can summary be any other type than string?

packs/vsphere/actions/workflows/vm_hw_basic_build.yaml
```diff
@@ -38,12 +39,13 @@ vsphere.vm_hw_basic_build:
...
         vsphere: <% $.vsphere %>
       publish:
-        vm_id: <% $.barebones_build.result.vm_id %>
+        vm_id: <% task(barebones_build).result.result.vm_id %>
       on-success:
         - add_controller
     add_controller:
       # [186, 198]
       action: vsphere.vm_hw_scsi_controller_add
+      wait-after: 12
```
Kami: Just curious - why are all those wait-after needed?

Are the operations async and we need to wait a bit before proceed with other operations? Or it it something else?

paul-mulvihill: they were added to spread the load when the workflow is called concurrently.
In my setup i can be calling it concurrently 10 times which on a single instance taxes mistral a lot. the waits allow things to run a bit more smoothly by giving actions time to complete in parrallel before the next step of the concurrent flows start.

packs/vsphere/actions/vm_env_items_get.py
paul-mulvihill: based on what pyvmomi returns it may be possible. the object is formated in some way which is kind of like json but isn't really. I hadn't updated that to convert to string as it still functions.
I can add a str conversion is that is prefered.

Kami: I see, thanks for the clarification.

Kami:
Looks good - merged, thanks!

PR #563, opened 2016-09-21
Improve GitHub pack with support for managing releases and deployments (+ add action aliases)

jjm:
### Status 

- pack extension.

### Description

Add actions using the GitHub APIs directly so that it's possible to manage GitHub releases & deployments via ST2 (and via ChatOPs).

- [X] Add Action Alias tests
- [x] Add support for connecting to GitHub.com and Enterprise at the same time.


packs/github/config.schema.yaml
```diff
@@ -0,0 +1,52 @@
+---
+token:
+  description: "GitHub oAuth Token"
+  type: "string"
+  default: "{{user.github_enterprise_oauth_token}}"
```
Kami: Nice, thanks!

packs/github/config.schema.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+user: 
+  description: "GitHub Username"
+  type: "string"
+
+password:
+  description: "GitHub Password"
+  type: "string"
+
+default:
```
Kami: Maybe `user_type` or `account_type` is a better name for this config option?

Kami: And it looks like in another place `github_type` parameter name is used so it's also good to be consistent and use the same name across the board :)

packs/github/actions/create_deployment_status.py
```diff
@@ -0,0 +1,68 @@
...
+import datetime
+
+from lib.base import BaseGithubAction
+
+
+class CreateDeploymentAction(BaseGithubAction):
+    def run(self, api_user, repository, deployment_id, state,
+            description, github_type):
+
+        valid_states = ["pending", "success", "error", "failure"]
```
Kami: Minor consistency thing - we tend to use use single quotes around strings.

packs/github/config.schema.yaml
jjm: Turns out I got this totally wrong :-( and forgot to correct it...

I've changed it to `system.github_oauth_token` as `user` scoping does not apply to action aliases.

packs/github/config.schema.yaml
jjm: Cool catch, I'll change it to `github_type` and only have _default_ in the description.

packs/github/actions/create_deployment_status.py
jjm: Cheers, fixed!

jjm:
Thanks for the review comments @Kami, these have been addressed and it's now ready to be merged if it looks ok to you too :-D

packs/github/config.schema.yaml
```diff
@@ -0,0 +1,54 @@
...
+  type: "string"
+
+password:
+  description: "GitHub Password"
+  type: "string"
+
+github_type:
+  description: "Default to either github or enterprise."
+  type: "string"
+  default: "online"
```
Kami: Maybe `public` or similar would be better than `online` but not a big deal anyway.

Kami:
Merged into master.

Nice works and thanks! :+1:

PR #565, opened 2016-09-29
New pack for Datadog

sanecz:
### Status 

- new pack, complete

### Description

Provides basic integration with the Datadog API accessible via their python library


packs/datadog/config.yaml
```diff
@@ -0,0 +1,3 @@
+---
+api_key: ''
```
Kami: When you get a chance, it would be great if you could also add config.schema.yaml file (support for config schema was added in a recent release of StackStorm) - https://docs.stackstorm.com/reference/pack_configs.html#configuration-schema

packs/datadog/pack.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
Kami: The icon should be 64x64 px, but not a big deal. I can resize it while merging.

packs/datadog/actions/run.py
```diff
@@ -0,0 +1,14 @@
+import importlib
+from st2actions.runners.pythonrunner import Action
+
+
+class ActionWrapper(Action):
+    def __init__(self, config):
+        self.config = config
+
+    def run(self, **kwargs):
```
Kami: Ideally you would explicitly declare all the supported arguments instead of using kwargs.

Kami: Also, just curious - were the metadata and action files some how auto-generated or not?

packs/datadog/requirements.txt
```diff
@@ -0,0 +1 @@
+datadog
```
Kami: Probably better to pin it to a specific version in case the API changes.

Kami:
Thanks, nice work!

There are some smaller things which could be improved (config schema, declaring arguments, etc.), but I will go ahead and merge what we have now.

It would be great if you could open another PR in the future with those changes / improvements.

packs/datadog/actions/run.py
sanecz: Yes, I started to write everything by hand but it took me a lot of time so I tried to auto-generate the biggest part but I had to write some things by hand. (I.e the functions taking an ID for editing or deleting things but that's not very hard, the type of the argument in the action file etc)

The main reason that I have not declared the supported arguments is because some arguments may be optional. If I set a default value, it will alter the behavior of the function. But I think I should declare only the required arguments and then the rest will be in the kwargs.

By the way, this method run is only a wrapper, so we need to keep the **kwargs here.

I'll still work on the pack for improving it, including unit tests too.

PR #566, opened 2016-09-30
added a change to print return value of function dev.system.chassis_n…

aniruddhamonker:
### Status

Bug fix
action vdx.system_chassis_name fails to return chassis name of the device.
### Description

Added change to vdx/actions/system_chassis.py to print the return the value function "dev.system.chassis_name(**kwargs)"


Kami:
Good catch, thanks.

Also, since this is a Python runner action any particular reason why a result is printed and not returned from the method?

Kami:
Actually, I just checked other actions and it looks like pretty much none of the existing actions print result of the function call.

If printing is desired (it makes debugging, etc.) easier, it would be good to update all of the existing actions.

/cc @bigmstone

bigmstone:
We should consider returning the value of the execution. I think that's what's desired here. I'll need to look into what's returned when there's no value returned. Perhaps if it's `NoneType` we can return 0 or similar.

Kami:
I do think that returning a value is a good idea but that needs more systematic approach (all the actions need to update, tests, etc.) so I will close this issue.

When we are going to tackle that we should open a new PR / issue.

PR #576, opened 2016-10-11
Added CLICRUD using the one true way

DavidJohnGee:
### Status 

CLICRUD enables simple CLI operations on Brocade ICX, MLX and VDX devices.
Two operations are supported: Single operational command and a list of configuration commands.

- New pack, tested and complete as a 1.0
- Want to add templating but will consult the ST2 team in terms of the right way of doing it

### Description

This is the first pull of CLICRUD and it works.


packs/clicrud/actions/config_command.yaml
```diff
@@ -0,0 +1,21 @@
...
+enabled: true
+entry_point: config_command.py
+parameters:
+  host:
+    type: string
+    description: Valid IPv4 address or hostname that resolves.
+    required: true
+  command:
+    type: array
+    description: List of commands separated by commas.
```
Kami: Just a heads up - "string separated by comma" is a CLI and UI convenience thing for supplying a list. if your provide parameter value via API or similar, you need to provide an actual JSON array.

packs/clicrud/actions/ops_command.py
```diff
@@ -0,0 +1,36 @@
...
+
+class OPSCommand(Action):
+
+    def run(self, host, command):
+        self.method = self.config['method']
+        self.username = self.config['username']
+        self.password = self.config['password']
+        self.enable = self.config['enable']
+
+        self.method = self.method.encode('utf-8', 'ignore')
```
Kami: I assume you call encode to make sure you we are operating with a string and not unicode? Or is there some other reason for it?

packs/clicrud/actions/ops_command.yaml
```diff
@@ -0,0 +1,15 @@
...
+enabled: true
+entry_point: ops_command.py
+parameters:
+  host:
+    type: string
+    description: Valid IPv4 address or hostname that resolves.
+    required: true
+  command:
+    type: string
+    description: Command for the operational 'show'
```
Kami: "operational" -> "operation" I assume :)

packs/clicrud/actions/config_command.py
```diff
@@ -0,0 +1,44 @@
...
+                                enable=self.enable, method=self.method,
+                                password=self.password)
+
+            return_value = transport.configure(utf8_commands)
+            _return_value = str(return_value).encode('utf-8', 'ignore')
+
+            if save is True:
+                transport.configure(["write mem"])
+
+            transport.close()
```
Kami: Is it saver to put `transport.close()` into `finally` block to make sure it's always executed even if some other line fails?

packs/clicrud/actions/ops_command.py
```diff
@@ -0,0 +1,36 @@
...
+            transport = generic(host=utf8_host, username=self.username,
+                                enable=self.enable, method=self.method,
+                                password=self.password)
+
+            return_value = transport.read(utf8_command, return_type="string")
+            return_value = unicode(return_value, "utf-8")
+            transport.close()
+            return return_value
+        except Exception, err:
+            self.logger.info('FUBARd')
```
Kami: I assume that's there for debugging purposes :)

packs/clicrud/config.yaml
```diff
@@ -0,0 +1,5 @@
+---
+  username: "admin"
```
Kami: If you get a chance it would also be great to add config.schema.yaml file - https://docs.stackstorm.com/reference/pack_configs.html

This way operator / admin can also specify password as in a datastore as a secret.

packs/clicrud/rules/gope_auto_remediate.yaml
```diff
@@ -0,0 +1,31 @@
+---
+description: 'GoPE remediate action for internet fail'
+tags: []
+type:
+  ref: standard
+  parameters:
+enabled: true
```
Kami: Rules are usually somewhat installation specific so the convention is to ship them disabled by default - this way operators can configure them first and then enable them.

Kami:
@DavidJohnGee How is this looking? I would appreciate explanation on utf8 / encoding stuff when you get a chance, but besides that and some minor typos PR looks good to me so I will go ahead and merge it (I will fix typos directly in master).

Thanks!

packs/clicrud/config.yaml
Kami: I pushed a change to master to replace config.yaml with config schema - 54d5534e3d437aed79f903b6327cd42103afa935

PR #579, opened 2016-10-15
Added support for 2nd Generation HUE bridges

tbraly:
### Status 

Bug Fix / Feature enhancement to support 2nd Generation Phillips HUE bridges

### Description

The Python-Hue library doesn't support the new authentication method used in the 2nd Generation Hue Bridge.  Easy work-around is to manually generate the API Key via the HUE Bridge WebGUI. Once configured in ST2, the library will use this key (if valid) and will not try to use 1st Generation authentication method.  The manual method will also work with 1st Generation. 

### DIFF
--- a/packs/hue/README.md
+++ b/packs/hue/README.md
@@ -8,12 +8,19 @@ This pack must be used on the same private network as the Hue bridge.
 ## Configuration
 
 * `station_ip` IP address or FQDN of Philips Hue Bridge
+* `client_identifier` API Key from the Philips Hue Bridge
 
 ## First Run
+
+### First Generation Hue Bridge
 
 The first run of any action will block until it is authorized. After
 installation, run the `hue.list_bulbs` and then press the button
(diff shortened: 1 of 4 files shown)


Kami:
Nice work, thanks!

Kami:
Do we also need to update `requirements.txt` or not (I assume changes depend on the new functionality in a client library)?

Edit: It looks like we don't pin to a specific version right now, would be a good idea to do it in any case.

tbraly:
The Phillips-Hue library has not been updated to support 2nd Gen Bridge Authentication, but it does support overriding of the client_identifier (aka. username) in the class. If the client_identifier is valid, then the 1st Generation method of authentication is never tried, thus allowing this method to work for both generations without a library modification.

Kami:
Ah, OK, thanks for the clarification and a heads up.

Works for me and looks good then!

Kami:
Merged, thanks.

PR #588, opened 2016-10-21
Config Schemas for all my packs

tonybaloney:
Pretty self-explanatory

### DIFF
--- a/packs/octopusdeploy/config.schema.yaml
+++ b/packs/octopusdeploy/config.schema.yaml
@@ -1,0 +1,16 @@
+---
+  api_key:
+    description: "API key"
+    type: "string"
+    secret: true
+    required: true
+  host:
+    description: "Octopus Deploy API host name"
+    type: "string"
+    secret: false
+    required: true
+  port:
+    description: "API region to use"
+    type: "string"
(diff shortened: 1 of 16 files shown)


Kami:
Nice!

Trying to get all those commits, huh :P

packs/octopusdeploy/config.schema.yaml
```diff
@@ -0,0 +1,16 @@
...
+    type: "string"
+    secret: true
+    required: true
+  host:
+    description: "Octopus Deploy API host name"
+    type: "string"
+    secret: false
+    required: true
+  port:
+    description: "API region to use"
```
Kami: Region -> port? :P

packs/powerpoint/config.schema.yaml
```diff
@@ -0,0 +1,6 @@
+---
+  template:
+    description: "Path to PPTX template file"
+    type: "string"
+    secret: false
```
Kami: False is a default value so if you prefer, you can leave `secret` attribute out all together for non-secret parameters.

Kami:
LGTM, +1

PR #594, opened 2016-10-24
Adding save_kittens pack

bigmstone:


armab:
A few days later ... 
http://petapixel.com/2014/03/04/malware-makes-use-of-sunset-and-cat-images-to-steal-banking-credentials/

bigmstone:
@armab Funny enough there's an image encoded with data as part of the story line. No malware though :D

PR #601, opened 2016-10-27
Pack validation changes

Kami:
This pull request updates pack metadata file so they will pass validation with new, to be released version of StackStorm.
## Background

Latest version of StackStorm changes some pack validation rules (version string now needs to be a valid `semver` identifier, email field needs to contain valid email address, etc.) and the pack.yaml is now also correctly validated the registration phase (previously there was a bug in the code so the pack metadata file wasn't actually validated).
## TODO
- [x] Add `ref` attribute with a value which matches directory name to each pack.yaml

### DIFF
--- a/packs/azure/pack.yaml
+++ b/packs/azure/pack.yaml
@@ -1,4 +1,5 @@
 ---
+ref: azure
 name : azure
 description : st2 content pack containing Microsoft Azure integrations.
 keywords:
--- a/packs/azure/pack.yaml
+++ b/packs/azure/pack.yaml
@@ -9,6 +9,6 @@ keywords:
   - servers
   - virtual machines
   - azure virtual machines
-version : 0.1
+version : 0.1.0
 author : StackStorm, Inc.
 email : info@stackstorm.com
--- a/packs/azure/pack.yaml
+++ b/packs/azure/pack.yaml
@@ -10,5 +10,5 @@ keywords:
   - virtual machines
   - azure virtual machines
 version : 0.1
-author : st2-dev
+author : StackStorm, Inc.
 email : info@stackstorm.com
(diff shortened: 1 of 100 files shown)


Kami:
I also confirmed this doesn't break current stable and older versions.

packs/consul/pack.yaml
```diff
@@ -1,8 +1,8 @@
 ### Place information about your pack version here.
 ---
+ref: consul
 name: consul
 description: consul
-version: 0.0.1
+version: 0.0.1.0
```
emedvedev: Oh.

packs/azure/pack.yaml
```diff
@@ -9,6 +10,6 @@ keywords:
   - servers
   - virtual machines
   - azure virtual machines
-version : 0.1
-author : st2-dev
-email : info@stackstorm.com
+version : 0.1.0
+author : StackStorm, Inc.
+email : info@stackstorm.com
```
Kami: Lol, yeah, just a personal preference.

I couldn't resist adding `.strip()` to the script which populated metadata files with ref field :D

packs/consul/pack.yaml
Kami: lol, yeah, that was a failed sed replacement :)

packs/azure/pack.yaml
Kami: @emedvedev I agree, I will push a change which adds line break back.

I actually only wanted to truncate additional empty lines at the end of the file (some files had them), but accidentally removed all the line breaks.

Kami: 96b098577c56fb1e873c3bd019455657b2206d57

PR #602, opened 2016-10-30
Added Sun Position Pack

tbraly:
### Status

New Pack, Complete
### Description

Provides a sensor that sends four triggers for dawn, sunrise, sunset, and dusk. 

Also provides actions for asking the current time (in UTC) of the for sun positions
for a specific geographical location.


Kami:
Nice.

Are you using that to figure out when you should get on the ground for a VFR flight? :smile:

packs/astral/actions/get_sunrise.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name: "get_sunrise"
+runner_type: "run-python"
+description: "Returns the time of the sunrise in UTC"
```
Kami: For convenience it would also be nice if user could supply a timezone / location (as a config option or action parameter or both).

packs/astral/config.schema.yaml
```diff
@@ -0,0 +1,11 @@
+---
+  latitude:
+    description: "Latitude of the location"
+    type: "integer"
```
Kami: Will probably need to use `number` here because integer doesn't allow decimal values iirc.

packs/astral/actions/get_sunrise.yaml
tbraly: Added Timezone support.  I looked into using location by city name, but there was some issues in the Astral code for city support. For example, Portland, Maine or Portland, Oregon. It had both as just "Portland".  Plus, I figured it's so easy these days to get your GEO, who cares about the city!

packs/astral/config.schema.yaml
tbraly: Fixed

tbraly:
Here in the USA, we are allowed to fly night VFR; however, I do agree that's not the best thing as mountains and clouds disappear at night.  

Right now I'm using it to turn on and off my Hue lights, but it will also take care of the Christmas lights next month.

tbraly:
I'm not getting the "Instance of 'GetSunriseAction' has no 'sun' member (no-member)" when I run pylint locally.  self.sun does exist as it's inherited from the BaseAction.  Any suggestions on how to get by this pylint error?

tbraly:
Mistakenly closed it

LindsayHill:
@tbraly may have to do something like this https://github.com/StackStorm/st2contrib/blob/6de75209fa4f1ae6b3d2d77e4e529e4d0298d971/packs/servicenow/actions/delete.py#L6

tbraly:
@LindsayHill Thanks, that did it!

packs/astral/actions/lib/BaseAction.py
```diff
@@ -0,0 +1,28 @@
+"""
```
Kami: Minor style thing - we use underscore separated names for modules, variable names, etc. Only use CamelCase for classes :)

packs/astral/actions/lib/BaseAction.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, config):
+        super(BaseAction, self).__init__(config)
+        self._latitude = self.config['latitude']
+        self._longitude = self.config['longitude']
+        self._timezone = self.config['timezone']
+
+        location = Location(('name', 'region', float(self._latitude),
+                            float(self._longitude), self._timezone, 0))
+
+        self.sun = location.sun()
```
Kami: > I'm not getting the "Instance of 'GetSunriseAction' has no 'sun' member (no-member)" when I run pylint locally. self.sun does exist as it's inherited from the BaseAction. Any suggestions on how to get by this pylint error?

Yeah sometimes pylint can't correctly infer all the variables exposed by a third party library (usually the problem is when a library does a lot of duck typing and other magic which makes it impossible to infer things during check / compile time).

packs/astral/actions/get_sunrise.yaml
Kami: Nice, thanks!

packs/astral/actions/lib/BaseAction.py
tbraly: made the fix to change BaseAction.py to action.py in just one commit 👍   Lesson learned that the GitHub web is not a good tool for making edit!

packs/astral/pack.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name : astral
+description : triggers for sunrise/sunset information
+version : 0.1
```
emedvedev: Version should be a valid semver string (i.e. `0.1.0`). That's a recent change, that's why you can still see `X.Y` versions in old packs, but there's a PR to fix this, and in all new packs we should use valid versioning from the start.

emedvedev:
@tbraly the pack icon should be centered, ideally. There's a couple cases where leaving the bottom half of the icon empty won't look good.

emedvedev:
Also, if anyone merges this PR without squashing it first, I'm going to be upset, leave a lot of passive-agressive comments, and haunt this person for life. 😛

LindsayHill:
Hey would you mind renaming `config.yaml` to `astral.yaml.example`? I'm trying to steer people away from using `config.yaml`, so have been adding `config.schema.yaml` as needed, and using `<pack_name>.yaml.example`, along with a note in README.md to tell people to edit & copy to `/opt/stackstorm/configs`. Goal is to get everyone onto the new method.

Kami:
Sorry for the delay - I will go ahead and merge this into master now.

Thanks again!

PR #608, opened 2016-11-04
Device42 get_dns_zone action

bearburger:
### Status 

- new action

### Description

New action `get_dns_zone` added.

### DIFF
--- a/packs/device42/actions/get_dns_zone.py
+++ b/packs/device42/actions/get_dns_zone.py
@@ -1,0 +1,37 @@
+import dns.rdataset
+import dns.zone
+
+from lib.base_action import BaseAction
+
+
+class GetDnsZone(BaseAction):
+    def run(self, domain, record_type=None, name=None, nameserver=None, content=None, tags=None,
+            tags_and=None):
+        response = self.getAPI("/api/1.0/dns/records/", {
+            "domain": domain,
+            "type": record_type,
+            "name": name,
+            "nameserver": nameserver,
+            "content": content,
+            "tags": tags,
+            "tags_and": tags_and
+        })
+
+        result = ""
+        if len(response["records"]) > 0:
+            zone = dns.zone.Zone(dns.name.from_text(response["records"][0]["dns_zone"]))
+            for record in response["records"]:
+                try:
+                    rdtype = dns.rdatatype.from_text(record["type"])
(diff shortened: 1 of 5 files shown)


packs/device42/actions/get_dns_zone.py
```diff
@@ -0,0 +1,37 @@
...
+                    n = zone.get_rdataset(record["name"], rdtype, create=True)
+                    n.add(rdata, record["ttl"])
+                except Exception, e:
+                    print "Error: %s" % e.message
+                    print "Record Name: %s" % record["name"]
+                    print "Record Type: %s" % record["type"]
+                    print "Record Content: %s" % record["content"]
+                    print "Record TTL: %d" % record["ttl"]
+                    print "---"
+            result = zone.to_text()
```
Kami: I believe it would be more useful if you returned structured data (a list of dictionaries instead of a text / string).

Returning a string means this data can't be simply used inside other actions and workflows (user would need to parse it, etc).

What do you think?

packs/device42/actions/get_dns_zone.py
```diff
@@ -0,0 +1,37 @@
...
+        })
+
+        result = ""
+        if len(response["records"]) > 0:
+            zone = dns.zone.Zone(dns.name.from_text(response["records"][0]["dns_zone"]))
+            for record in response["records"]:
+                try:
+                    rdtype = dns.rdatatype.from_text(record["type"])
+                    rdata = dns.rdata.from_text(dns.rdataclass.IN, rdtype, record["content"])
+                    n = zone.get_rdataset(record["name"], rdtype, create=True)
```
Kami: Hm, I'm a bit confused why this line or the loop itself is needed - it doesn't really seem it's used or am I missing something?

Kami:
@bearburger Thanks, I've added some comments.

packs/device42/actions/get_dns_zone.py
Kami: Ah, in this case I'm fine with it, but I still see value in a separate "get_dns_records` action or similar which returns a structured list of records.

Kami:
I'll go ahead and merge this into master and we can work on separate `get_dns_records` or similar action in the future if needed :)

bearburger:
Hi @Kami Sorry, I was away. I agree with you, in some cases list might be useful. I'll add this action in next few days.

PR #609, opened 2016-11-04
New actions for consul pack

TuxInvader:
### Status 

Add actions to register and de-register external services, and also update the query_service action to return IPs and ports. 

- pack extension

### Description

query_service has an additional ports=<boolean> parameter, which defaults to false. When true, the returned list includes the port as ip:port.

New action: register_service can be used to register an external node/service with consul.
New action: deregister_service. Removes a registered node/service from consul.

### DIFF
--- a/packs/consul/actions/register_service.py
+++ b/packs/consul/actions/register_service.py
@@ -2,8 +2,8 @@ from lib import action
 
 
 class ConsulRegisterServiceAction(action.ConsulBaseAction):
-    def run(self, nid, name, address, port, tags, dc):
+    def run(self, node, service, address, port, tags, dc):
 
-        definition = {"Service": name, "Port": port, "Tags": tags}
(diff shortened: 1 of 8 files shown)


Kami:
Thanks - changes look good to me.

Kami:
Merged - thanks!

packs/consul/actions/register_service.py
```diff
@@ -0,0 +1,9 @@
+from lib import action
+
+
+class ConsulRegisterServiceAction(action.ConsulBaseAction):
+    def run(self, node, service, address, port, tags, dc):
+
+        definition = {"Service": service, "Port": port, "Tags": tags}
+        result = self.consul.catalog.register(node, address, service=definition, dc=dc)
+        return (result, "")
```
Kami: Actually, I missed this earlier - any particular reason you return a tuple?

You could simply just return a result.

PR #610, opened 2016-11-12
Microsoft Excel Pack

tbraly:
### Status 

new pack, complete

### Description

This pack is for reading and writing to Excel Spreadsheets as a information store.

This can come in handy when you want to grab data for workflows and/or store
results from workflows in a format that is easily editable and formatable.


Kami:
Thanks.

I think you and @tonybaloney should be friends - Anthony wrote a PowerPoint pack (my condolences :trollface:) :P

tonybaloney:
This is awesome. Now we just need Microsoft Word and BWC is feature complete.

packs/excel/config.schema.yaml
```diff
@@ -0,0 +1,30 @@
+---
+  excel_file:
```
Kami: One option would be to make this a parameter of each action.

This way user doesn't have to update config file, etc. if they are working with multiple spreadsheets. A good middle ground between a config parameter and action parameter would be to make action parameter optional and use config file if action parameter is not provided, otherwise use the parameter (this way we would support both approaches and it would be easy to work with multiple spreadsheets).

Kami:
Thanks again - I've checked it out locally and played with it and besides the comments I added, things look good :+1:

It would be great if we could at least address missing init file and excel_file as an optional parameter thing. Moving to lockfile or similar would also be good, but not a blocker for niw :)

packs/excel/actions/get_keys.py
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from lib import excel_action, excel_reader
+
+
+class GetExcelSheetsAction(excel_action.ExcelAction):
+    def run(self, sheet):
+        excel = excel_reader.ExcelReader(self._excel_file)  # pylint: disable=no-member
```
Kami: And once that's added, let's please also remove pylint ignore since they shouldn't be needed anymore and this time it looks like they caught an actual issue :)

packs/excel/actions/get_keys.yaml
```diff
@@ -0,0 +1,11 @@
+---
+name: "get_keys"
+runner_type: "run-python"
+description: "returns an array of the keys for a given sheet in the excel file"
+enabled: true
+entry_point: "get_keys.py"
```
Kami: I tested it locally and it looks like this action returns first column from each of the rows.

I think it would also make sense to allow user to retrieve all the columns in the first row as well - usually first line is the header which contains all the keys. Or perhaps we should add new action (e.g. `get_keys_for_rows`) and rename this one `get_keys_for_columns` or similar?

packs/excel/actions/lib/excel_reader.py
```diff
@@ -0,0 +1,239 @@
...
+        sheets = []
+        for sheet in self._wb:
+            sheets.append(sheet.title)
+        return sheets
+
+    def get_keys(self):
+        ''' Returns the keys in the sheet '''
+        if not self._ws:
+            raise KeyError("Sheet not specified")
+        keys = []
```
sanecz: Maybe `keys = self._keys[:]` is a more pythonic solution

packs/excel/actions/lib/excel_reader.py
```diff
@@ -0,0 +1,239 @@
...
+    def save(self):
+        ''' Saves spreadsheet to disk '''
+        if not self._lock:
+            raise IOError("Trying to save file not locked for modification")
+        self._wb.save(self._filename)
+        self._unlock_file()
+
+    def get_sheets(self):
+        ''' Returns an array of the sheet names '''
+        sheets = []
```
sanecz: You should probably use a list comprehension `sheets = [sheet.title for sheet in self._wb]`

packs/excel/actions/get_keys.py
tbraly: That would explain why I wasn't getting the error locally. Forgot to add __init__.py on the push

packs/excel/config.schema.yaml
tbraly: Given each spreadsheet could have different starting rows/columns, I'll add those as optional parameters as well.

packs/excel/actions/get_keys.yaml
tbraly: Let me do that

packs/excel/actions/lib/excel_reader.py
tbraly:
```
    keys = self._keys[:]
TypeError: unhashable type
```

but

`list(self._keys)` works

packs/excel/actions/lib/excel_reader.py
```diff
@@ -0,0 +1,239 @@
...
+            # is file already locked?
+            attempts = 0
+            while os.path.isfile("%s.lock" % excel_file):
+                time.sleep(lock_delay)
+                attempts += 1
+                if attempts >= lock_retries:
+                    raise IOError("File is locked")
+            # it's now ours, so lock it
+            try:
+                open("%s.lock" % excel_file, 'a').close()
```
tbraly: It says the "lockfile" has be depreciated

packs/excel/actions/lib/excel_reader.py
tbraly: made the change

packs/excel/actions/lib/excel_reader.py
Kami: Ah, you are right - might be worth looking in the other two suggested packages although StackStorm code and a lot of other projects I know still use `lockfile` so it's probably also not a big deal if we still use it :)

packs/excel/config.schema.yaml
Kami: Thanks.

Let me know when you have pushed all the changes and I will have a look.

packs/excel/actions/lib/excel_reader.py
```diff
@@ -0,0 +1,248 @@
...
+        self._variable_start_column = key_column + 1
+        self._variable_end_column = self._variable_start_column
+        self._set_variable_start_column(key_column + 1)
+
+    def get_row_for_key(self, key):
+        ''' Returns the row for a given key, or -1 if not matched '''
+        if len(key) > 255:
+            self._unlock_file()
+            raise ValueError("Key exceeds 255 characters")
+        if key in self._keys:
```
sanecz: I 'll probably go for a `return self._keys.get(key)` which will return None if key is not found and then in set_values_for_variables, test if row is not None instead of testing if row equals -1

Kami:
I confirmed that the new changes you made work locally (thanks!), made two small changes / improvements (d312bad96437e5180af896544e805db82ded428a, 872adf355be32a8c987193716c11e688d2d38c9c) and merged PR into master.

Thanks again.

PR #616, opened 2016-11-17
Support DNS round-robin balancing through Route53

emedvedev:
Our codegen actions for adding/updating A records in Route53 only support single IP as a value. This PR makes them accept a comma-separated list, which will add an unweighted round-robin A record with multiple IPs.

Should also add WRR at some point probably, but I don't care nearly enough to do that.

### DIFF
--- a/packs/aws/actions/run.py
+++ b/packs/aws/actions/run.py
@@ -12,6 +12,8 @@ class ActionManager(action.BaseAction):
             kwargs['user_data'] = self.st2_user_data()
         if action == 'create_tags':
             kwargs['tags'] = self.split_tags(kwargs['tags'])
+        if action in ('add_a', 'update_a'):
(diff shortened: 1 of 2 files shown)


tonybaloney:
Why not just use the libcloud DNS code? then at least you're not bound to R53

emedvedev:
@tonybaloney I wasn't sure it would support round-robin (took a brief look at the docs and the code, but apparently need to dig deeper), and it was pretty easy to fix in the AWS pack since boto2 does support it for sure.

Overall, in production, especially with mixed environments, I would use libcloud for most things, but I'm still not ready to abandon the AWS pack because it'll feel neglected and sad :cry: There's also some value for the demo in particular: the less layers of abstraction to focus on, the better.

Kami:
LGTM :+1:

PR #617, opened 2016-11-17
Validate example pack configs against schema during builds.

Kami:
The title says it all.

This pull request depends on the st2 code change in https://github.com/StackStorm/st2/pull/3033.


Kami:
Decided to move the utility script into st2sdk so it can be used in other places as well - https://github.com/StackStorm/st2sdk/pull/8.

Kami:
Fixed all the offending configs and schemas in ab87bfb.

LindsayHill:
Do you think we should pick up these sorts of mistakes too? https://github.com/StackStorm/st2contrib/pull/622/commits/73ec0272440a0774f5a0dc92e65b42544c83767a

PR #619, opened 2016-11-18
HPE ICSP pack rename

paul-mulvihill:
### Status 
- pack extension

### Description

Pack Rename to remove '-' and replace with '_'
Fixes #619 

### Checklist (tick everything that applies)

- [X] Metadata: pack.yaml, icon, structure (required for new packs)
- [X] Version bump (required for changed packs)
- [X] Code linting (required, can be done after the PR checks)
- [ ] [Tests](https://docs.stackstorm.com/development/pack_testing.html) (not required but really recommended)

PR #620, opened 2016-11-18
hpe_icsp migrate to config.schema

LindsayHill:
### Status 

- pack extension, complete

### Description

Migrate hpe_icsp to `config.schema.yaml`


LindsayHill:
@paul-mulvihill - Thanks for the change in #619 - I've further updated it to move to `config.schema.yaml`

LindsayHill:
Hmm, looks like the new pack checks weren't run on my local box. Will need to check my local setup and fix these issues

Kami:
It looks like you just need to update the sample config and specify sample values of the right type for all the required config options.

LindsayHill:
Yep, looks that way. Trying to work out why my local system complaining
about missing imports when running make configs-check. Part of the problem
is working on the train with intermittent connectivity.

Will fix it when I get to the office

Kami:
Yeah - you need latest changes from st2 repo and st2sdk for the make targets to work - running make configs-check should indeed pull down those dependencies but yeah, bad connectivity could cause issues.

LindsayHill:
And now tests pass...had to do some manual imports on my local system to run checks locally. Not too worried about that right now though.

Kami:
LGTM :+1:

PR #622, opened 2016-11-18
Fix missing init, Removed pylint overides

tbraly:
### Status 

- bug fixes

### Description

Fix missing init, Removed pylint overides


LindsayHill:
If you're making changes to this pack, can you also fix up the typos in `config.schema.yaml`? Just noticed that it needs s/requried/required/

Kami:
Thanks and good catch @LindsayHill.

In theory we could set "additionalProperties" to false in st2 code base so it would throw on "required" typo, but would need to dig in and see if and how if affects existing stuff.

Issue #623, opened 2016-11-19
[AWS] Couldn't get result of Buckets in S3


userlocalhost2000:
Hi,

The result of actions that return Bucket, e,g `aws.s3_get_bucket` and `aws.s3_get_all_buckets`, couldn't return proper value. They always return `None`.

I confirmed this behavior just only for the actions that return `Bucket` object of S3.
The actions relevant to EC2 return proper results.

Thank you.

userlocalhost2000:
The reason of this problem is failure of serialization of the returned values.

I'll fix it :)

ErikThoreson:
@userlocalhost after this PR it seems certain actions receive the following error "AttributeError: 'module' object has no attribute 'bucket'"

Were you able to get around this? It looks like a problem with boto 2.41.0 and higher

userlocalhost:
@ErikThoreson  Thank you for your report.

I could get buckets information normally.
```
vagrant@st2-node:~$ st2 run aws.s3_get_all_buckets
.
id: 584bc0226f086f0572eacaf4
status: succeeded
parameters: None
result: 
  exit_code: 0
  result:
  - LoggingGroup: http://acs.amazonaws.com/groups/s3/LogDelivery
    creation_date: '2016-11-15T10:45:45.000Z'
    name: 20161115-tmp01
  - LoggingGroup: http://acs.amazonaws.com/groups/s3/LogDelivery
    creation_date: '2016-11-15T11:01:05.000Z'
    name: 20161115-tmp02
  - LoggingGroup: http://acs.amazonaws.com/groups/s3/LogDelivery
    creation_date: '2015-11-20T02:58:47.000Z'
    name: dmzo
  - LoggingGroup: http://acs.amazonaws.com/groups/s3/LogDelivery
    creation_date: '2016-11-14T06:58:25.000Z'
    name: hoge-20161114-02
  stderr: ''
  stdout: ''
vagrant@st2-node:~$
```

And this is my environment of StackStorm.
```
vagrant@st2-node:~$ st2 --version
st2 2.1.0
vagrant@st2-node:~$ /opt/stackstorm/virtualenvs/aws/bin/pip show boto
Name: boto
Version: 2.41.0
Summary: Amazon Web Services Library
Home-page: https://github.com/boto/boto/
Author: Mitch Garnaat
Author-email: mitch@garnaat.com
License: MIT
Location: /opt/stackstorm/virtualenvs/aws/lib/python2.7/site-packages
Requires: 
[...]

PR #624, opened 2016-11-19
the correction of #623

userlocalhost2000:
### Status 

- bugfix

### Description
The main point of this correction is
* adding a condition to serialize s3.bucket.Bucket in the type selector.
* removing the `connection` field from returning Bucket object, because the `connection` field has a python object which is inherited `AWSAuthConnection`. If the returned result has python object, parse processing in the `PythonRunner` may fail. And the action result will be `None` as reported in #623.

- [x] [Tests](https://docs.stackstorm.com/development/pack_testing.html) (not required but really recommended)

### DIFF
--- a/packs/aws/tests/fixtures/full.yaml
+++ b/packs/aws/tests/fixtures/full.yaml
@@ -1,0 +1,14 @@
+---
+aws_access_key_id: "access_key_id"
+aws_secret_access_key: "secret_key"
+region: "us-west-1"
+interval: 20
+st2_user_data: ""
+
+service_notifications_sensor:
+  host: "localhost"
+  port: 12345
+  path: "/my-path"
+
+sqs_sensor:
+  input_queues: "input_queue"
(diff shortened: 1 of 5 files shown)


Kami:
Thanks - the change looks good to me, but it looks like the tests are failing - https://travis-ci.org/StackStorm/st2contrib/builds/177250896

packs/aws/tests/fixtures/full.yaml
```diff
@@ -0,0 +1,14 @@
+---
+aws_access_key_id: "access_key_id"
+aws_secret_access_key: "secret_key"
+region: "us-west-1"
+interval: 20
+st2_user_data: "/opt/stackstorm/packs/aws/actions/scripts/bootstrap_user.sh"
```
Kami: Looks like the tests are failing because this config option is specified.

Simply set it to an empty string or none and the tests should pass I believe.

userlocalhost2000: Ops, thank you for your advice. I'll fix it.

Kami:
Merged, thanks!

PR #625, opened 2016-11-21
migrate newrelic to config.schema

LindsayHill:
### Status 

- pack extension, complete

### Description

migrate to `config.schema.yaml`


packs/newrelic/config.schema.yaml
```diff
@@ -0,0 +1,38 @@
...
+      host:
+        type: "string"
+        description: "IP Address sensor should listen on. Default 0.0.0.0"
+        required: false
+        default: "0.0.0.0"
+      port:
+        type: "integer"
+        description: "Port Address sensor should listen on. Default 10001"
+        required: false
+        default: 10001
```
Kami: This reminds me - I actually need to check / verify that default values work with nested objects.

Kami: Yeah, so I just checked / verified and we don't support default values for nested objects right now.

Because of that, we sadly need to wait with migration of packs which use nested config properties with default values.

LindsayHill:
Thanks Kami, I'll make some fixes, so it's ready to merge once the other
changes are in

On Mon, Nov 21, 2016 at 6:16 AM Tomaz Muraus <notifications@github.com>
wrote:

*@Kami* commented on this pull request.
------------------------------

In packs/newrelic/config.schema.yaml
<https://github.com/StackStorm/st2contrib/pull/625>:

> +  sensor_config:
+    description: "Sensor specific settings."
+    type: "object"
+    required: true
+    additionalProperties: false
+    properties:
+      host:
+        type: "string"
+        description: "IP Address sensor should listen on. Default 0.0.0.0"
+        required: false
+        default: "0.0.0.0"
+      port:
+        type: "integer"
+        description: "Port Address sensor should listen on. Default 10001"
+        required: false
+        default: 10001

We will be able to merge this once StackStorm/st2#3039
<https://github.com/StackStorm/st2/pull/3039> is merged into master.

Just make sure to set required: false for the top-level object attribute (
sensor_config) or leave this attribute out completely since it defaults to
false.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/StackStorm/st2contrib/pull/625>, or mute the thread

Kami:
@LindsayHill No problem.

And we should be backward and forward compatible as long as we set `require: false` to any nested object attribute.

Now that those changes are in, we can also re-consider if it still makes sense to flatten the puppet pack or not.

LindsayHill:
> Now that those changes are in, we can also re-consider if it still makes sense to flatten the puppet pack or not.
I'm inclined to still make the Puppet flattening change, since it's a relatively simple config, and doesn't gain anything from nesting.

Kami:
LGTM

PR #626, opened 2016-11-21
flattened puppet config and moved to config.schema

LindsayHill:
### Status 

- pack extension, complete

### Description

Flattened configuration and moved to `config.schema.yaml`. NB this will be breaking change for existing configs.

### DIFF
--- a/packs/puppet/CHANGES.md
+++ b/packs/puppet/CHANGES.md
@@ -2,7 +2,15 @@
 
 ## v0.2.0
 
-Flattened configuration and migrated to config.schema.yaml
+**BREAKING CHANGE**
+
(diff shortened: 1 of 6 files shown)


Kami:
Thanks.

The change looks good to me, but flatting is backward-incompatible and we use puppet pack ourselves so we need to be careful we make it clear that backward incompatible change was made in readme and similar.

LindsayHill:
Yeah, I thought about that later. I'll add some more detail to README and
CHANGES

Kami:
Thanks for updating the readme and changelog.

It looks good to me now, we just need to be careful and monitor support channels for mentions of config related issues when installing / upgrading the pack so we can point them to the right direction.

PR #627, opened 2016-11-21
migrate mmonit to config.schema

LindsayHill:
### Status 

- pack extension, complete

### Description

Migrated to `config.schema.yaml`

### DIFF
--- a/packs/mmonit/config.schema.yaml
+++ b/packs/mmonit/config.schema.yaml
@@ -1,0 +1,44 @@
+---
+  host:
+    description: "Mmonit host to connect to for actions."
+    type: "string"
+    required: true
+    secret: false
+  username:
+    description: "Username to connect to mmonit for actions"
+    type: "string"
+    required: true
+    secret: false
+  password:
+    description: "Password to connect to mmonit for actions"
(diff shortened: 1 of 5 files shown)


packs/mmonit/config.schema.yaml
```diff
@@ -0,0 +1,44 @@
...
+      username:
+        description: "Username to connect to mmonit for sensor"
+        type: "string"
+        required: true
+      password:
+        description: "Password to connect to mmonit for sensor"
+        type: "string"
+        required: true
+      event_types:
+        description: "Comma separated list of events to listen for. Possible values: 0=success, 1=error, 2=change. 1 and 2 by default"
```
Kami: Ideally this should be an array and the code should convert it to that format but that's a backward incompatible change we can do some time later.

Kami:
Merged, thanks.

packs/mmonit/config.schema.yaml
Kami: I decided to do it anyway in #632.

LindsayHill:
Thanks!

Issue #629, opened 2016-11-22
[AWS] AWSSQSSensor doesn't refer to top-level authentication parameters in the configuration file.


userlocalhost2000:
Hi, 

AWSSQSSensor may have following problems.

* It doesn't refer to the top-level credential parameters, only do them in `setup` directory which is remained for backword compatibility.
* ~~It refers to the configuration parameter of 'sqs_other.max_number_of_messages', but the default configuration doesn't have it~~.

userlocalhost2000:
Sorry, the later was my misunderstanding.

userlocalhost2000:
This correction is merged.

Thank you!

PR #631, opened 2016-11-23
correcting the problem which is mentioned in #629

userlocalhost2000:
### Status
- bugfix

### Description

This patch add a processing to get values which are defined in top-level on the configuration.

In the past, the top-level parameters on the configuration (like `aws_access_*` and `region`) have never been referred. Instead, other params which are specified under the `setup` dictionary have been done.

This patch corrects to get these configuration parameters in the following order.

1. from the datastore
2. from configuration that is under the dictionary which is specified in `prefix` argument. 
3. from configuration that is under the `setup` directory for backwards compatibility.

This way enables to refer params which is define in arbitrary place as well as have backwards compatibility using the same interface.

Thank you.

- [x] [Tests](https://docs.stackstorm.com/development/pack_testing.html) (not required but really recommended)

### DIFF
--- a/packs/aws/pack.yaml
+++ b/packs/aws/pack.yaml
@@ -18,6 +18,6 @@ keywords:
   - RDS
   - SQS
 
-version : 0.7.1
+version : 0.7.2
 author : StackStorm, Inc.
 email : info@stackstorm.com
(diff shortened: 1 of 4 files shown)


userlocalhost2000:
I've finished this work. Could you please merge this?

emedvedev:
@userlocalhost2000 thank you, this is a great contribution! Sorry I couldn't review earlier. Merging.

PR #650, opened 2017-08-30
[DO NOT MERGE] BREAKING st2contrib deprecation

Mierdin:
> **PLEASE READ**: this PR is not intended to be merged to `master`. Please read on for the proposal

StackStorm 2.1 originally introduced a brand new way of working with packs. As a result, this repository was deprecated in favor of the [StackStorm Exchange](https://github.com/StackStorm-exchange/). However, the analytics still show that this repository is still used in some capacity.

We want to eventually get to a place where we can remove the content here, but the main driver is twofold:

- Prevent Google from indexing this content, so that users searching for packs hit the Exchange repos before this one
- Lead users to updated content that probably has much more recent activity and bugfixes

In lieu of just deleting this repository to achieve this goal, we recognize that some organizations may have just not had a chance to upgrade to a recent version yet. So, we're deciding to move all content to an `archive` directory. This is technically a breaking change, but the content is still available for those that absolutely must use it as it exists in `st2contrib`. We encourage all users that encounter this to seriously consider upgrading to a recent version of StackStorm

In addition to the directory change, I'm proposing that we don't merge this branch into `master`; instead, we designate this branch as the primary branch for this repository, and delete the `master` branch. This will prevent Google from indexing this content.

I'll keep this PR open for a few days before doing anything further, but the above proposal will go into effect soon if I don't hear compelling objections.


LindsayHill:
I am in favor of this proposal. It's important that the content is still there if someone really needs it. But the key is that they should really need it, and know what they're doing, not just stumble across it. 

I would like us to update as many stackstorm blogs as possible, to update st2contrib references to point to new pack locations

Kami:
If we set some other non-master branch as a default one, Google won't index it? Didn't know that.

I'm +1, but we should also try to find all the existing posts, articles, etc. and update the links where we can.

Mierdin:
Let's decide to make the change on Tuesday Sep 5th. Don't want to do it at the end of the week or over the labor day weekend. That should give us more than enough time to update blogs.

decided

arm4b:
👎 on removing, moving or deleting content here.

There is one big reason: it will break all StackStorm versions trying to install the packs before st2 `2.1`. 1yr old is not that big time if we talk about enterprises or not and there must be some users having that version freezed.
I think old st2 installations is one of the reasons why we still have traffic here.

I think placing REALLY BIG DEPRECATION NOTICE in every pack `readme.md` would help a bit, but I don't like the idea moving packs to `archive` subdir because it will break any automated deployments users had codified.

Mierdin:
@armab IMO, if users haven't seen the blogs and deprecation notices we have posted by now, they won't see any others. I agree that this will break automation deployments pre-2.1, that's the point. Moving to `archive` means the break will get their attention, but not all will be lost. They can make the choice to amend their automation, or upgrade. This is a "shot across the bow".

Remember, the goal is to lead folks to exchange packs. Having this content on the internet compromises that, due to Google indexing. The alternative is removing it altogether - and the branch change and directory move is middle ground - gets us the benefit of not confusing new users with two locations for packs, and also doesn't leave old installations totally without recourse.

So this just comes down to "Is a year long enough to allow users to upgrade?" IMO, as easy as stackstorm is to upgrade, I think a year is more than enough time, and the tradeoff against the potential benefits (new users being able to find updated content more easily) is worth it.

arm4b:
Updating README can _help_ to prevent some contributions in this repo and redirect them to new repos, - I agree it won't do anything else.

Talking about the breaking change for older st2 installations.
As a user of hundreds software tools, I can expect breaking changes in new versions, but I'd wish to burn in fire any software that suddenly breaks existing already working stuff for me and I suspect majority would prefer the same :)

arm4b:
If we're doing this and move to subdir, can we at least verify how updated pack installation will look like on old st2 versions?
For instance, this will be broken:
```
st2 run packs.install packs=github,aws
```
https://docs.stackstorm.com/2.0/packs.html#getting-a-pack
How the fixed format will look like to install these 2 packs on older st2 (with subdir)?

bigmstone:
My primary issue w/ `st2contrib` is that it takes precedence in search results. If we change default branch away from `master` but leave the packs in place I'm not sure if that will break existing installs of ST2.

The good thing about this change is that it doesn't out-right break ST2. Only pack installs. We should have a clear way to fix it for people if they encounter it - with strong encouragement to upgrade. To say in another way you have to actively touch/use ST2 and specifically pack install to see this issue. It's not like someone's CI is going to break when they wake up.

bigmstone:
Also I _think_ we still have servers in our infra that will be affected by this. 😂

LindsayHill:
Biggest thing is the Google indexing. Wish that was as easy to stop as adding robots.txt

arm4b:
+1 just a branch rename looks safe way to solve the Google indexing (without the `archive` subdir breaking change).

We just need to test it beforehand.

nmaludy:
https://webapps.stackexchange.com/questions/67344/how-to-prevent-a-public-github-repository-from-showing-up-on-search-results

Edit: read that wrong. @armab  looks right

Mierdin:
Okay, so definitely the indexing problem is the bigger of the two, but there's also the problem of existing users using old pack content. I know there have been a ton of fixes (both functional and security) to a number of the packs that I use, and that's just a fraction of them. The purpose of changing the location of these files is to get users' attention, and let them know of this fact. Changing the indexing problem fixes things for new users, but I feel like it's a good time to make a more drastic notification.

I'll just change the default branch tomorrow (to a new, third branch that's not this one), since we don't have consensus on the `archive` directory change, but I'd like us to seriously considering doing it in the near future. Perhaps step up the "deprecation notice" game in the next month or two, then use the directory change to get attention and point to those notices.

Mierdin:
So I spun up a `2.0.1` instance of st2, and even changing the default branch will break things, since the default `branch` parameter is `master`:

```
vagrant@st2vagrant:/opt/stackstorm/packs/packs/actions/pack_mgmt$ st2 run packs.install packs=twitter repo_url="Mierdin/st2contrib"
..
id: 59b0834dc4da5f21159201ca
action.ref: packs.install
parameters:
  packs:
  - twitter
  repo_url: Mierdin/st2contrib
status: failed
error: Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/runners/python_action_wrapper.py", line 211, in <module>
    obj.run()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/runners/python_action_wrapper.py", line 126, in run
    output = action.run(**self._parameters)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 94, in run
    branch=branch)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 123, in _clone_repo
    Repo.clone_from(repo_url, abs_local_path, branch=branch, depth=1)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/repo/base.py", line 743, in clone_from
    return cls._clone(Git(os.getcwd()), url, to_path, GitCmdObjectDB, progress, **kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/repo/base.py", line 695, in _clone
    finalize_process(proc)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/util.py", line 132, in finalize_process
    proc.wait()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/cmd.py", line 112, in wait
    raise GitCommandError(self.args, status, self.proc.stderr.read())
git.exc.GitCommandError: 'git clone --depth=1 --branch=master -v https://github.com/Mierdin/st2contrib.git /root/st2contrib' returned exit status 128: Cloning into '/root/st2contrib'...
warning: Could not find remote branch master to clone.
fatal: Remote branch master not found in upstream origin
Unexpected end of command stream


traceback: None
failed_on: download
start_timestamp: 2017-09-06T23:22:53.248681Z
end_timestamp: 2017-09-06T23:22:55.549240Z
+--------------------------+---------------------+----------+----------------+-------------------------------+
| id                       | status              | task     | action         | start_timestamp               |
+--------------------------+---------------------+----------+----------------+-------------------------------+
| 59b0834dc4da5f1d770a7b7b | failed (1s elapsed) | download | packs.download | Wed, 06 Sep 2017 23:22:53 UTC |
+--------------------------+---------------------+----------+----------------+-------------------------------+
vagrant@st2vagrant:/opt/stackstorm/packs/packs/actions/pack_mgmt$ st2 run packs.install packs=twitter repo_url="Mierdin/st2contrib" branch="deprecation-test"
..........
id: 59b08354c4da5f21159201cd
action.ref: packs.install
parameters:
  branch: deprecation-test
  packs:
  - twitter
  repo_url: Mierdin/st2contrib
status: succeeded
start_timestamp: 2017-09-06T23:23:00.455693Z
end_timestamp: 2017-09-06T23:23:19.298412Z
+--------------------------+------------------------+--------------------------+-------------------------+-------------------------------+
| id                       | status                 | task                     | action                  | start_timestamp               |
+--------------------------+------------------------+--------------------------+-------------------------+-------------------------------+
| 59b08354c4da5f1d770a7b7e | succeeded (4s elapsed) | download                 | packs.download          | Wed, 06 Sep 2017 23:23:00 UTC |
| 59b08358c4da5f1d770a7b80 | succeeded (2s elapsed) | virtualenv_prerun        | packs.virtualenv_prerun | Wed, 06 Sep 2017 23:23:04 UTC |
| 59b0835ac4da5f1d770a7b83 | succeeded (6s elapsed) | setup_virtualenv         | packs.setup_virtualenv  | Wed, 06 Sep 2017 23:23:06 UTC |
| 59b08361c4da5f1d770a7b85 | succeeded (4s elapsed) | reload                   | packs.load              | Wed, 06 Sep 2017 23:23:13 UTC |
| 59b08366c4da5f1d770a7b88 | succeeded (0s elapsed) | restart-sensor-container | packs.restart_component | Wed, 06 Sep 2017 23:23:18 UTC |
+--------------------------+------------------------+--------------------------+-------------------------+-------------------------------+
```

While this can of course be overridden via parameters, it's still considered a breaking change, since anyone with automated workflows that are installing packs on this version will probably need to add this parameter, else their pack installs/updates will break.

In addition, it's probably good form for us to first consider the update of all of the pack READMEs and blog articles. Let's first tackle the removal of all `st2contrib` references from everything, and add urgent warnings where relevant, indicating that an upgrade to 2.1 or later is a good idea, and very soon, things will start breaking. Then once done, we can start down the breaking path, like changing the default branch and ultimately, removal of st2contrib

